PR #11, opened 2015-02-05
Improve tests landing page

MattiSG:
Avoid misplaced access requests.


jdesboeufs:
Pareil, mais trop tard, ce code est mort !

PR #18, opened 2015-04-09
Remove compass dependency

fpagnoux:


jdesboeufs:
Thx!

package.json
```diff
@@ -62,7 +62,7 @@
     "time-grunt": "~1.1.0"
   },
   "engines": {
-    "node": ">=0.10.0"
+    "node": "0.10"
```
MattiSG: This could have been on its own commit  :)

MattiSG:
Great! Now we just have to update `ludwig-ui` and we can drop Ruby as a dependency  :tada:

PR #43, opened 2015-04-20
Update mes-aides-api to v0.1.4

MattiSG:
Should fix mongoose dependency.
Poke @Flightan.


Flightan:
This does indeed fix mongoose dependency.

PR #51, opened 2015-04-24
Import tests

fpagnoux:
To import the test from prod in a local mongo database


MattiSG:
Good to merge.

MattiSG:
Whoops, no. We need to document this in the README.

PR #57, opened 2015-04-26
accessibility fix

goetsu:
still need to generate aria-describedby="id-of-corresponding-help-block" on input when inputs have help-block


MattiSG:
Thanks @goetsu. We have a deadline coming up next week, so I'm unsure about the priorisation of this PR and #56, as we need to add the dynamic `describedby`. I'll try to handle it if I have some slack, otherwise we'll probably see it next week.

goetsu:
as you want but there is no harm to already apply this because it containt other fixes (correct for/id mechanism and aria-describedby for group of multiple inputs)

MattiSG:
Merged via cherry-pick in 0cb8033537918c99486c7636b72f96de164b1185. Thanks again!  :)

PR #71, opened 2015-05-05
Fix bug creation test and cerfa

fpagnoux:


MattiSG:
Have you tested the Cerfa generation with and without this change?

fpagnoux:
Not, I haven't tested Cerfa

MattiSG:
Ok, I'll do it.

MattiSG:
Indeed, there's no link to the Cerfas anymore… Good catch. I'm now double-checking all other links to Cerfa.

We really need integration tests.

MattiSG:
Fixed via #73 (cherry-picked this PR).

PR #136, opened 2015-08-12
Ne pas afficher les prestations capturées dans les résultats

fpagnoux:
Connected to #118 

Todo : 

- [x] Test
- [x] Componentize the html part


app/js/services/simulationService.js
```diff
@@ -1,54 +1,37 @@
...
-    var filterIneligible = makeObjectFilter(function(value, key) {
-        return value;
-    });
-
-    function describe(droits) {
-        return _.mapValues(droits, function(montant, key) {
-            var result = _.clone(droitsDescription[key]);
-            result.montant = montant;
-            return result;
+    // Si la valeur renvoyée par l'API est nulle, cela signifie par convention que l'aide a été injectée et non recaculée par le simulateur
```
MattiSG: Ambiguité : "nulle" = `null` ou 0 ?

app/js/services/simulationService.js
```diff
@@ -1,54 +1,37 @@
...
+    function sortDroits(droitsCalcules) {
+        var droitsEligibles = {};
+        var droitsInjectees = {};
+        var droitsNonEligibles = {};
```
MattiSG: Use a single `var` with commas for grouped declarations.

app/js/services/simulationService.js
```diff
@@ -1,54 +1,37 @@
...
+
+        _.forEach(droitsDescription, function(description, droitKey) {
+            if (droitsCalcules[droitKey]) {
+                droitsEligibles[droitKey] = description;
+                droitsEligibles[droitKey].montant = droitsCalcules[droitKey];
+            } else if (droitsCalcules[droitKey] === null) {
+                droitsInjectees[droitKey] = description;
+            }
+            else {
```
MattiSG: Inconsistent newlines. Always put operators on the same line as curly braces.

app/js/services/simulationService.js
```diff
@@ -1,54 +1,37 @@
...
+                droitsNonEligibles[droitKey] = description;
+            }
         });
+        return {
+            'droitsEligibles' : droitsEligibles,
```
MattiSG: No need for quotes in object keys.

app/js/services/simulationService.js
```diff
@@ -1,54 +1,37 @@
...
+            'droitsInjectees' : droitsInjectees,
```
MattiSG: No space before colon.

app/views/partials/droits-list.html
```diff
@@ -0,0 +1,27 @@
...
+            class="btn btn-sm btn-default"
+            target="_blank">
+            <i class="fa fa-external-link"></i>
+            Plus d'informations
+          </a>
+        </accordion-group>
+      </div>
+    </div>
+  </div>
+</accordion>
```
MattiSG: Missing newline at EoF. Use an editor plugin to support the [EditorConfig](http://editorconfig.org) setup we have.

app/views/partials/simulation.html
```diff
@@ -134,35 +134,11 @@
...
-              <i class="fa fa-external-link"></i>
-              Plus d'informations
-            </a>
-          </accordion-group>
-        </div>
-      </div>
-    </div>
-  </accordion>
-</div>
+  <droits-list droits="droitsNonEligibles"></droits-list>
```
MattiSG: :+1:

test/spec/services/simulationService.js
```diff
@@ -56,16 +15,20 @@ describe('SimulationService', function () {
...
-            var actual = service.complement({ acs: { montant: 12 } });
-
-            expect(actual.acs).toBe(undefined);
-            expect(actual.apl.id).toBe('apl');
+        it('should sort eligible and non eligible droits', function() {
+            var openfiscaResult = { acs: 10 };
+            var droits = service.sortDroits(openfiscaResult);
+            expect(droits.droitsEligibles).toEqual({ acs: { id: 'acs', montant: 10 } });
+            expect(droits.droitsNonEligibles).toEqual({ apl: { id: 'apl' } });
+            expect(droits.droitsInjectees).toEqual({});
```
MattiSG: Whenever possible, have each `expect` in its own `it` and group them within a `describe`, if necessary with a `before`. That makes for shorter, more specific tests that are more precise when they fail and better support evolution of the codebase.

app/js/services/simulationService.js
fpagnoux: Ok, mais par curiosité quel est l'avantage à grouper les déclarations de variables ?

test/spec/services/simulationService.js
fpagnoux: So it would be better to have one big expect about droits instead of 3 ones ?
Because i don't think it would be relevant to split the 3 expects, there are not independent.

app/views/partials/droits-list.html
fpagnoux: Ok, thanks !

test/spec/services/simulationService.js
MattiSG: Yes, it would be better because the generated messages would be clearer.

If, for example, the algorithm that computes `droitsNonEligibles` fails, it is much clearer to have two messages that end with “…should compute droits non eligibles” under two different `describe` rather than knowing that both “should sort eligible and non eligible droits” and “should sort eligible and injected droits” fail. In the latter case, you have to find the lines and read the code to pinpoint the failure, while it is obvious otherwise.

You can also notice that the title of your tests is misleading: it states “it('should sort eligible and non eligible droits'”, yet you test `expect(droits.droitsInjectees).toEqual({})`. If that last line fails, the message will be `SimulationService sortDroits should sort eligible and non eligible droits failed: expected {a:1} to equal {}`, while that very sorting of eligible and non eligible will be fine.

app/js/services/simulationService.js
MattiSG: Avantage, [baisse de risque](https://jslinterrors.com/combine-this-with-the-previous-var-statement) ou [inconvénient](http://stackoverflow.com/questions/694102/declaring-multiple-variables-in-javascript), c'est surtout un [choix](http://programmers.stackexchange.com/questions/229694/multiple-var-statements-in-javascript)  ;)  Personnellement, je préfère car ça enlève du bruit visuel.

fpagnoux:
Comments treated :)

app/js/filters/isEmpty.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+angular.module('ddsCommon')
+    .filter('isEmpty', function() {
+        return function(object) {
+            return _.isEmpty(object);
```
MattiSG: :+1:

app/js/services/simulationService.js
```diff
@@ -1,54 +1,36 @@
...
-
-    function describe(droits) {
-        return _.mapValues(droits, function(montant, key) {
-            var result = _.clone(droitsDescription[key]);
-            result.montant = montant;
-            return result;
+    // Si la valeur renvoyée par l'API vaut null, cela signifie par convention que l'aide a été injectée et non recaculée par le simulateur
+    function sortDroits(droitsCalcules) {
+        var droitsEligibles = {},
+        droitsInjectes = {},
```
MattiSG: Indent.

app/views/partials/simulation.html
```diff
@@ -134,35 +134,11 @@
 
 <div>
   <hr>
+  <h4 ng-show="!(droitsInjectes|isEmpty)">Aides que vous percevez déjà aujourd'hui, et dont nous n'avons pas recalculé le montant</h4>
```
MattiSG: spaces around operators (unary `!` and binary `|`).

PR #145, opened 2015-08-25
Improve income capture on last 12 months

fpagnoux:
Small improvements before resources page big bang. 
Connected to #22

### DIFF
--- a/app/styles/front.scss
+++ b/app/styles/front.scss
@@ -306,17 +306,15 @@ $autre-lighter: #cce4e7;
   border-left-width: 4px;
   margin-bottom: 10px;
 
-  h2 {
+  .capture-title {
     margin: 0;
-    padding: 0;
     padding-left: 10px;
     font-weight: 500;
-  }
-
-  .capture-title {
-    padding: 5px 0;
+    font-size: 21px;
     background-color: lighten($color-light, 10%);
     color: #333;
+    cursor: pointer;
+    user-select: none;
   }
 
   .capture-rows {
--- a/app/views/partials/foyer/ressources/types.html
+++ b/app/views/partials/foyer/ressources/types.html
@@ -1,7 +1,7 @@
 <h1>{{ pageTitle }}</h1>
 
 <p>
-  Sélectionnez les ressources perçues <strong>entre {{ debutAnneeGlissante }} et {{ finAnneeGlissante }}</strong>,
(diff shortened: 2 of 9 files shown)


app/views/partials/foyer/ressources/types.html
```diff
@@ -1,16 +1,15 @@
...
 </p>
 
 <form method="post" name="form" novalidate ng-submit="submit()">
   <div class="ressource-types-block">
     <div class="capture-{{ category.id }}" ng-repeat="category in ressourceCategories">
-      <div class="capture-title">
-        <h2>{{ category.label }}</h2>
-      </div>
+      <details>
+        <summary class="capture-title">{{ category.label }}</summary>
```
MattiSG: `<details>` needs to be [polyfilled](https://en.wikipedia.org/wiki/Polyfill) for IE11. If you want to use it, add [this one](https://github.com/MattiSG/Element.details/tree/mattisg).

app/views/partials/foyer/ressources/types.html
```diff
@@ -1,16 +1,15 @@
...
       <div class="capture-rows">
```
MattiSG: You changed nesting, update indentation.

app/styles/front.scss
```diff
@@ -306,17 +306,15 @@ $autre-lighter: #cce4e7;
   border-left-width: 4px;
   margin-bottom: 10px;
 
-  h2 {
```
MattiSG: Changing instructions for such a generic selector is extremely dangerous. Have you visited _every instance of `<h2>` throughout the app_ after having made the change?

fpagnoux: Does this h2 block apply to all h2 throughout the app ? It's inside `@mixin ressource-category-capture` block, so I thought it only applied on a restricted area.

MattiSG: You’re right, I missed the two-spaces indentation in GitHub's diff view.

app/views/partials/foyer/ressources/types.html
fpagnoux: Polyfication didn't work :tired_face:

fpagnoux: I gave a last shot to bootstrap's `accordion` but it would need big changes on the style sheet that may be not worth it.

PR #150, opened 2015-09-04
Engager les démarches pour Paris Logement Famille

fpagnoux:
Connected to #131


app/views/partials/resultat.html
```diff
@@ -84,6 +84,14 @@
               Plus d'informations
             </a>
           </div>
+          <div ng-if="droit.id == 'paris_logement_familles'"
+            class="col-sm-3 col-sm-offset-6">
+            <a target="_self"
+              class="btn btn-sm btn-primary btn-block"
+              href="/documents/Formulaire_demande_PLF.pdf" download>
```
fpagnoux: @MattiSG,
I know `download` doesn't work for all browsers, but do you have any idea to let the bénéficiaire potentiel access the pdf file without leaving the `mes-aides` world ?

MattiSG: Yes: don't force to download in the current presentation. The current button has too many different meanings, so let's just open it in another tab with `target="_blank"`. If the user has a PDF reader, it will stay in the browser while keeping the simulation open, and if not, it will download it.

fpagnoux: Ok done !

MattiSG:
GTM

MattiSG:
No route is opened for the newly-created folder `documents`, the document is not served in production mode.

fpagnoux:
That should be fixed, thanks @Flightan

MattiSG:
GTM!!

PR #160, opened 2015-09-17
Implement resources screen with bootsrap ui

fpagnoux:
Make it work on Firefox on IE

Praise @Flightan :bow: :bow: :bow:

### DIFF
--- a/app/views/partials/foyer/ressources/types.html
+++ b/app/views/partials/foyer/ressources/types.html
@@ -7,9 +7,11 @@
 
 <form method="post" name="form" novalidate ng-submit="submit()">
   <div class="ressource-types-block">
-    <div class="capture-{{ category.id }}" ng-repeat="category in ressourceCategories">
-      <details>
-        <summary class="capture-title">{{ category.label }}</summary>
+    <accordion>
+      <accordion-group class="capture-{{ category.id }}" ng-repeat="category in ressourceCategories" is-open="status.open">
+        <accordion-heading>
+          {{ category.label }}<i class="pull-left fa" ng-class="{'fa-chevron-down': status.open, 'fa-chevron-right': !status.open}"></i>
+        </accordion-heading>
         <div class="capture-rows">
           <button
             type="button"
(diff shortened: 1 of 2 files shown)


fpagnoux:
image

app/views/partials/foyer/ressources/types.html
```diff
@@ -5,23 +5,26 @@
...
+            <div class="capture-rows">
+              <button
+                type="button"
+                class="btn btn-sm btn-selection btn-default"
+                ng-repeat="ressourceType in ressourceTypesByCategories[category.id]"
+                btn-checkbox
+                ng-model="individuVM.selectedRessourceTypes[ressourceType.id]">
+                {{ ressourceType.label }}
+              </button>
+            </div>  </accordion-group>
```
MattiSG: Line break.

app/views/partials/foyer/ressources/types.html
```diff
@@ -5,23 +5,26 @@
   vous pourrez ensuite saisir les montants.
 </p>
 
+
+
```
MattiSG: Why these additional line breaks?

fpagnoux:
Thanks for your feedback ! Is this good now ?

MattiSG:
Should be, I just need to test it manually…

MattiSG:
I'm really sorry, but I don't think this behaviour is proper. Only one item can be open at the time, which means I have no clue what I have already clicked on. The current behaviour to keep open tabs open is perhaps not the best as it allows one to clutter the screen, but this one is not acceptable: if I toggle a resource on, it is pretty hard to toggle it back off: where was it? What have I already selected?

One workaround could perhaps be to add the number of toggled resource types in the header (not great, but a workaround).

MattiSG:
(FYI I have already looked at the [Bootstrap doc](http://getbootstrap.com/javascript/index.html#collapse-options) and there doesn't seem to be a way to allow multiple selections on an accordion)

fpagnoux:
There must be a way to dis-activate the "one at a time" behavior. Looking at it right now.

fpagnoux:
It seems to be possible to avoid this behavior. I'm on it.
https://angular-ui.github.io/bootstrap/#/accordion

fpagnoux:
That should work. @MattiSG, is it ok on your browser too ?

MattiSG:
GTM!!

PR #163, opened 2015-09-23
Appliquer correctement la majoration du RSA pour isolement

fpagnoux:
Connected to #156

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -16,5 +16,6 @@
     "catch",
     "while"
   ],
-  "requireCamelCaseOrUpperCaseIdentifiers": "ignoreProperties"
+  "requireCamelCaseOrUpperCaseIdentifiers": "ignoreProperties",
+  "requireSpaceAfterPrefixUnaryOperators": ["++", "--", "~", "!"]
 }
--- a/.jscsrc
+++ b/.jscsrc
@@ -15,5 +15,6 @@
     "try",
     "catch",
     "while"
-  ]
+  ],
+  "requireCamelCaseOrUpperCaseIdentifiers": "ignoreProperties"
 }
--- a/app/js/controllers/foyer.js
+++ b/app/js/controllers/foyer.js
@@ -17,7 +17,7 @@ angular.module('ddsApp').controller('FoyerCtrl', function($scope, $state, $state
         } else {
             situation.individus.push(demandeur);
         }
-        $state.go('foyer.conjoint');
+        $state.go('foyer.personnesACharge');
     });
 
     $scope.$on('individu.conjoint', function(e, conjoint) {
@@ -33,7 +33,7 @@ angular.module('ddsApp').controller('FoyerCtrl', function($scope, $state, $state
                 return 'conjoint' !== individu.role;
             });
         }
-        $state.go('foyer.personnesACharge');
+        $state.go('foyer.logement');
     });
 
     $scope.$on('personnesACharge', function(e, personnesACharge) {
@@ -42,7 +42,7 @@ angular.module('ddsApp').controller('FoyerCtrl', function($scope, $state, $state
         });
 
         situation.individus = situation.individus.concat(personnesACharge);
-        $state.go('foyer.logement');
+        $state.go('foyer.conjoint');
     });
 
     var isLogementInMayotte = function(logement) {
--- a/app/js/controllers/foyer.js
+++ b/app/js/controllers/foyer.js
@@ -32,8 +32,10 @@ angular.module('ddsApp').controller('FoyerCtrl', function($scope, $state, $state
             situation.individus = _.filter(situation.individus, function(individu) {
                 return 'conjoint' !== individu.role;
             });
+        } // En cas de parent isolé, on pose une question supplémentaire
+        if (! SituationService.hasEnfant($scope.situation)) {
+            $state.go('foyer.logement');
         }
-        $state.go('foyer.logement');
     });
 
     $scope.$on('personnesACharge', function(e, personnesACharge) {
--- a/app/js/controllers/foyer.js
+++ b/app/js/controllers/foyer.js
@@ -32,7 +32,8 @@ angular.module('ddsApp').controller('FoyerCtrl', function($scope, $state, $state
             situation.individus = _.filter(situation.individus, function(individu) {
                 return 'conjoint' !== individu.role;
             });
-        } // En cas de parent isolé, on pose une question supplémentaire
+        }
+        // En cas de parent isolé, on pose une question supplémentaire
         if (! SituationService.hasEnfant($scope.situation)) {
             $state.go('foyer.logement');
         }
--- a/app/js/controllers/foyer/conjoint.js
+++ b/app/js/controllers/foyer/conjoint.js
@@ -1,6 +1,17 @@
 'use strict';
 
-angular.module('ddsApp').controller('FoyerConjointCtrl', function($scope) {
+angular.module('ddsApp').controller('FoyerConjointCtrl', function($scope, $state, SituationService) {
+    var hasEnfant = SituationService.hasEnfant($scope.situation);
+    $scope.demandeur = SituationService.getDemandeur($scope.situation);
+
+    $scope.captureIsolement = function() {
+        return angular.isDefined($scope.vitEnCouple) && (! $scope.vitEnCouple) && hasEnfant;
+    };
+
+    $scope.submit = function() {
+        $state.go('foyer.logement');
+    }
+
     if (_.find($scope.situation.individus, { role: 'conjoint' })) {
         $scope.vitEnCouple = true;
     }
--- a/app/js/controllers/foyer/conjoint.js
+++ b/app/js/controllers/foyer/conjoint.js
@@ -5,7 +5,7 @@ angular.module('ddsApp').controller('FoyerConjointCtrl', function($scope, $state
     $scope.demandeur = SituationService.getDemandeur($scope.situation);
 
     $scope.captureIsolement = function() {
-        return angular.isDefined($scope.vitEnCouple) && (! $scope.vitEnCouple) && hasEnfant;
+        return $scope.vitEnCouple === false && hasEnfant;
     };
 
     $scope.submit = function() {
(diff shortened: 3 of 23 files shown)


app/js/controllers/foyer.js
```diff
@@ -32,8 +32,10 @@ angular.module('ddsApp').controller('FoyerCtrl', function($scope, $state, $state
             situation.individus = _.filter(situation.individus, function(individu) {
                 return 'conjoint' !== individu.role;
             });
+        } // En cas de parent isolé, on pose une question supplémentaire
```
MattiSG: This comment is linked to the line below. It should either be appended at the end of the line it describes, or on its own line right above it. Here, it looks like it explains the closing curly brace.

app/js/controllers/foyer/conjoint.js
```diff
@@ -1,6 +1,17 @@
 'use strict';
 
-angular.module('ddsApp').controller('FoyerConjointCtrl', function($scope) {
+angular.module('ddsApp').controller('FoyerConjointCtrl', function($scope, $state, SituationService) {
+    var hasEnfant = SituationService.hasEnfant($scope.situation);
+    $scope.demandeur = SituationService.getDemandeur($scope.situation);
+
+    $scope.captureIsolement = function() {
+        return angular.isDefined($scope.vitEnCouple) && (! $scope.vitEnCouple) && hasEnfant;
```
MattiSG: Why both `isDefined` and `!`? Don't you simply want `=== false`?

app/js/services/situationService.js
```diff
@@ -187,8 +187,16 @@ angular.module('ddsApp').factory('SituationService', function($http, $sessionSto
...
+        getDemandeur: function(situation) {
+            return _.find(situation.individus, { role: 'demandeur' });
+        },
+
         hasEnfantScolarise: function(situation) {
             return !!_.find(situation.individus, { role: 'enfant', scolarite: 'college' }) || !!_.find(situation.individus, { role: 'enfant', scolarite: 'lycee' });
+        },
+
+        hasEnfant: function(situation) {
+            return !!_.find(situation.individus, { role: 'enfant' });
```
MattiSG: Space after unary operators (one may consider double-bang construct as a unary operator).

In general, avoid that kind of idiomatic-but-unreadable casts. If you really do need to cast to a `Boolean`, use `Boolean(val)` rather than `!! val` (additionally, the higher cost of writing will make you think twice before doing it  ;)  ).

app/js/controllers/foyer.js
fpagnoux: Done.

app/js/services/situationService.js
fpagnoux: I'm adding the space rule to the .jscsrc.

fpagnoux: So in JS, as long as the code is fine with it, you shouldn't cast ? 
The fact that the function name `hasEnfant` suggests it returns a `boolean` is not considered a problem ?

app/js/controllers/foyer/conjoint.js
fpagnoux: Indeed.
Thanks :smiley:

app/js/services/situationService.js
MattiSG: > So in JS, as long as the code is fine with it, you shouldn't cast ?

Well, it is a matter of consistency, of course. If your codebase relies on `=== true`, you should obviously cast. On the other hand however, if you do take advantage of truthies and falsies and write `if (hasEnfant)`, what's the point of casting it? Wasting (admittedly very little) CPU and memory?

As soon as you understand that _everything_ in JS can be evaluated to a boolean value, there is little need for actual `Boolean` types, except in domain representations and specific cases where you would have a different treatment for different falsies (such as `undefined` ≠ `false`, as we've seen above).
This doesn't mean you should _avoid_ `Boolean` values. It simply means `return myVal` is as much a boolean value as `return !! myVal`. The fact that the function name suggests it returns a `true`/`false` value actually (to me) suggests it is very safe to return truthies/falsies, as its return value would be only used in a comparison or a condition.

Then again, consistency above all. But if you want to return a `Boolean`, I'd go with a clear, readable cast where even a JS newbie understands what goes on rather than the cryptic `!!`.

.jscsrc
```diff
@@ -15,5 +15,7 @@
     "try",
     "catch",
     "while"
-  ]
+  ],
+  "requireCamelCaseOrUpperCaseIdentifiers": "ignoreProperties",
+  "requireSpaceAfterPrefixUnaryOperators": ["++", "--", "~", "!"]
```
MattiSG: :+1:

app/js/controllers/foyer/conjoint.js
```diff
@@ -1,6 +1,17 @@
 'use strict';
 
-angular.module('ddsApp').controller('FoyerConjointCtrl', function($scope) {
+angular.module('ddsApp').controller('FoyerConjointCtrl', function($scope, $state, SituationService) {
+    var hasEnfant = SituationService.hasEnfant($scope.situation);
+    $scope.demandeur = SituationService.getDemandeur($scope.situation);
+
+    $scope.captureIsolement = function() {
+        return $scope.vitEnCouple === false && (! $scope.vitEnCouple) && hasEnfant;
```
MattiSG: The `(! $scope.vitEnCouple)` is redundant with `=== false` ;)

app/js/controllers/foyer/individuForm.js
```diff
@@ -48,7 +48,7 @@ angular.module('ddsApp').controller('FoyerIndividuFormCtrl', function($scope, in
         role: individuRole
     };
 
-    if (true === ($scope.captureRelationConjoint = !!options.captureRelationConjoint)) {
+    if (true === ($scope.captureRelationConjoint = !! options.captureRelationConjoint)) {
```
MattiSG: Just to keep the conversation on truthies/falsies going on, and not asking for changes here as you obviously just fixed whitespace (thanks!), this is an excellent example of added complexity. This could simply be rewritten to:
```js
$scope.captureRelationConjoint = options.captureRelationConjoint;
if ($scope.captureRelationConjoint) {
```

app/js/services/situationService.js
```diff
@@ -187,8 +187,16 @@ angular.module('ddsApp').factory('SituationService', function($http, $sessionSto
 
         flattenPatrimoine: flattenPatrimoine,
 
+        getDemandeur: function(situation) {
+            return _.find(situation.individus, { role: 'demandeur' });
+        },
+
         hasEnfantScolarise: function(situation) {
-            return !!_.find(situation.individus, { role: 'enfant', scolarite: 'college' }) || !!_.find(situation.individus, { role: 'enfant', scolarite: 'lycee' });
+            return _.find(situation.individus, { role: 'enfant', scolarite: 'college' }) || !! _.find(situation.individus, { role: 'enfant', scolarite: 'lycee' });
```
MattiSG: Cast in second operand  ;)

app/views/partials/foyer/conjoint.html
```diff
@@ -6,14 +6,39 @@
         class="btn btn-default btn-block"
         btn-radio="false"
         ng-model="vitEnCouple"
-        focus-on="vitSeul"
-        ng-click="$emit('individu.conjoint', null)">Je vis seul(e)</button>
+        ng-click="$emit('individu.conjoint', null)">
+        Je vis seul(e)
```
MattiSG: Unify masculine/feminine distinctions to dots rather than parens (`.e`), more readable.

app/views/partials/foyer/conjoint.html
```diff
@@ -6,14 +6,39 @@
...
+  <div ng-if ="captureIsolement()">
+  <h2>Depuis combien de temps vivez-vous seul(e) ?</h2>
+  <div class="row">
+    <div class="col-md-4 col-md-offset-2">
+      <button
+        class="btn btn-default btn-block"
+        btn-radio="false"
+        ng-model="demandeur.isolementRecent"
+        ng-change="submit()">
+        Plus de 18 mois
```
MattiSG: Unify buttons with reading direction: “Moins de 18 mois” before “Plus de 18 mois”.

test/spec/services/situationService.js
```diff
@@ -195,7 +195,41 @@ describe('Service: situationService', function () {
             });
 
             // then
-            expect(results).toEqual([true, true, false]);
+            expect(results[0]).toBeTruthy();
+            expect(results[1]).toBeTruthy();
+            expect(results[2]).toBeFalsy();
+        });
+    });
+    describe('function hasEnfant()', function() {
```
MattiSG: Great, congrats on expanding coverage!  :)

app/js/controllers/foyer/individuForm.js
fpagnoux: Yep, I remember seeing one of those and getting really upset ;)

app/js/controllers/foyer/conjoint.js
fpagnoux: Done

app/js/services/situationService.js
fpagnoux: Done

app/views/partials/foyer/conjoint.html
fpagnoux: Pour `demandeur`,  `demandeu.r.se` ?
Je suis pas sûr pour la lisibilité...

app/views/partials/foyer/conjoint.html
fpagnoux: Done

fpagnoux:
Corrections done :)

fpagnoux:
Corrections done, but the behavior is flawed... I'll need some help on this one

fpagnoux:
Problem solved ! 
@MattiSG , is it GTM ?

app/js/controllers/foyer/conjoint.js
```diff
@@ -1,6 +1,20 @@
 'use strict';
 
-angular.module('ddsApp').controller('FoyerConjointCtrl', function($scope) {
+angular.module('ddsApp').controller('FoyerConjointCtrl', function($scope, $state, SituationService) {
+    var hasEnfant = SituationService.hasEnfant($scope.situation);
+    $scope.demandeur = SituationService.getDemandeur($scope.situation);
+
+    // Forget the value of isolementRecent. Necessary as we une ng-change to go to the next page.
```
MattiSG: “use”

MattiSG:
GTM

PR #174, opened 2015-10-05
Refactor tns resources

fpagnoux:
Depends on 
Connected to sgmap/openfisca-france#12

### DIFF
--- a/app/js/constants/ressources.js
+++ b/app/js/constants/ressources.js
@@ -201,7 +201,8 @@ angular.module('ddsCommon').constant('ressourceTypes', [
     {
         id: 'caMicroEntreprise',
         label: 'Micro-entreprise',
-        category: 'rpns'
+        category: 'rpns',
+        isMontantAnnuel: true
     },
     {
         id: 'caAutoEntrepreneur',
@@ -211,12 +212,14 @@ angular.module('ddsCommon').constant('ressourceTypes', [
     {
         id: 'revenusAgricolesTns',
         label: 'Exploitant agricole',
-        category: 'rpns'
+        category: 'rpns',
+        isMontantAnnuel: true
     },
     {
         id: 'autresRevenusTns',
         label: 'Profession libérale, entrepreneur',
-        category: 'rpns'
+        category: 'rpns',
+        isMontantAnnuel: true
     }
 ]);
 
--- a/app/js/controllers/foyer/recapSituation.js
+++ b/app/js/controllers/foyer/recapSituation.js
@@ -160,8 +160,7 @@ angular.module('ddsCommon').controller('RecapSituationCtrl', function($scope, $s
                     type: ressourceType,
                     totalAnnuel: Math.round($scope.tempRessources[ressourceType.id].totalAnnuel)
                 };
-                var ressourcesWithoutMonthlyAmount = ['caMicroEntreprise','revenusAgricolesTns','autresRevenusTns'];
-                if (! _.contains(ressourcesWithoutMonthlyAmount, ressourceType.id)) {
+                if (! ressourceType.isMontantAnnuel) {
                     ressource.totalMensuel = $scope.tempRessources[ressourceType.id].totalMensuel;
                 }
                 $scope.ressources.push(ressource);
--- a/app/js/controllers/foyer/recapSituation.js
+++ b/app/js/controllers/foyer/recapSituation.js
@@ -144,8 +144,6 @@ angular.module('ddsCommon').controller('RecapSituationCtrl', function($scope, $s
     var buildRecapRessources = function() {
         $scope.tempRessources = {};
         $scope.hasRessources = false;
-        $scope.hasRessourcesTns = false;
-        $scope.hasRessourcesNonTns = false;
         $scope.globalAmount = 0;
 
         $scope.isSituationMonoIndividu = 1 === $scope.situation.individus.length;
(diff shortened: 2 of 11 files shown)


app/js/controllers/foyer/ressources/ressources.js
```diff
@@ -67,53 +63,23 @@ angular.module('ddsApp').controller('FoyerRessourcesCtrl', function($scope, $sta
                 onGoing: true
             };
 
+            // For autres revenus TNS, we also need to find the CA
+            if (type == 'autresRevenusTns') {
+                ressource.caAnnuel = _.chain(ressources)
+                    .where({ type: 'caAutresRevenusTns' })
+                    .pluck('montant')
+                    .reduce(function(sum, num) {
```
MattiSG: `num` is badly named, as it is ambiguous with an index. Use either `montant` for consistency with the `pluck` if you consider it a part of the domain, or `amount` otherwise.

app/js/controllers/foyer/ressources/ressources.js
```diff
@@ -135,31 +101,14 @@ angular.module('ddsApp').controller('FoyerRessourcesCtrl', function($scope, $sta
...
-                };
-
-                _.each(individuKeysToRessourceKeys[ressource.type.id], function(ressourceKey, individuKey) {
-                    individu[individuKey] = ressource[ressourceKey];
-                });
-            }
 
+        individuVM.ressources.forEach(function(ressource) {
+            // Ressources for which we have the last 3 months values
             if (ressource.type.category != 'rpns' || ressource.type.id == 'caAutoEntrepreneur') {
```
MattiSG: > It is always better to branch the less frequent cases in the `if` part, and the most frequent in the `else` branch (or, even better, to use an early exit in the `if` and remove the `else` altogether). That makes for less branches to keep open in your mental model of the data flow when reading the code.

(Not very important, so no request to change.)

test/spec/controllers/foyer/ressources/ressources.js
```diff
@@ -276,7 +262,7 @@ describe('Controller: FoyerRessourcesCtrl', function() {
             expect(individu.interruptedRessources[0]).toEqual('test2');
         });
 
-        it('devrait sauvegarder les "autres revenus professionnels" directement sur l’individu', function() {
+        it('devrait sauvegarder les "autres revenus professionnels" dans les ressources', function() {
```
MattiSG: English consistency would be welcome.

Not mandatory here, though, but it would follow the [boy-scout rule](http://programmer.97things.oreilly.com/wiki/index.php/The_Boy_Scout_Rule)  :wink:

test/spec/services/ressourceService.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
+
+describe('ResultatService', function () {
+    describe('applyYearlyRessource', function() {
+        var service, individu, dateDeValeur, ressource;
```
MattiSG: Put each `var` declaration on its own line. Less code churn and more readability.

MattiSG:
So much deleted code, yet increased quality coverage. Congratulations, that was beautiful  :)

The code seems GTM as is, but I didn't double-check behaviour, and would rather avoid spending time on this.

Yet, this is quite a major change. What do you think about deploying to `next` and letting @Anne-LaureM validate usage? That would definitely make me more confident.

fpagnoux:
I agree, there is a lot of changes so only thorough test would let us make sure nothing is broken. I'll deploy it on next when we are done with the other validation.

app/js/controllers/foyer/ressources/ressources.js
fpagnoux: Done

test/spec/controllers/foyer/ressources/ressources.js
fpagnoux: Done

test/spec/services/ressourceService.js
fpagnoux: Done

MattiSG:
GTM on code. Waiting for validation.

fpagnoux:
The situation recap was not great, I just updated it to fit with changes in the model. Tell me if it looks good to you (last commit), i'm deploying on demo for validation.

app/views/partials/foyer/recap-situation.html
```diff
@@ -21,33 +21,21 @@ <h4 class="section-header">Vos ressources</h4>
...
-        </div>
-        <div class="col-sm-4">
-          {{ ressource.totalMensuel[$index] }} €
+      <div ng-if="ressource.totalMensuel">
+        <div class="row" ng-repeat="month in months">
+          <div class="col-sm-8">
+            {{ month.label | uppercaseFirst }}
+          </div>
+          <div class="col-sm-4">
+            {{ ressource.totalMensuel[$index] }} €
```
MattiSG: Mmh. Am I supposed to see an amount?
capture d ecran 2015-10-13 a 09 29 02

app/js/controllers/foyer/recapSituation.js
```diff
@@ -155,56 +153,20 @@ angular.module('ddsCommon').controller('RecapSituationCtrl', function($scope, $s
...
-            }
-            if (individu.autresRevenusTns) {
-                ressourcesAutresTns.totalAnnuel += individu.autresRevenusTns;
-                $scope.hasRessources = true;
-            }
-           if (individu.revenusAgricolesTns) {
-                ressourceAgricoleTns.totalAnnuel += individu.revenusAgricolesTns;
-                $scope.hasRessources = true;
+                };
+                var ressourcesWithoutMonthlyAmount = ['caMicroEntreprise','revenusAgricolesTns','autresRevenusTns'];
```
MattiSG: Good refactor.

This part should be a constant, and should be initialised outside from the loop. Here, for each resource, you are allocating a new array and three new strings; since the name is not uppercased, it also feels like it is mutable, while it isn't.

Actually, shouldn't this be something like `_.find(Resources, { isMontantAnnuel: true })`? (pseudo-code, I don't know `_`'s syntax for this)

app/js/controllers/foyer/recapSituation.js
```diff
@@ -155,56 +153,20 @@ angular.module('ddsCommon').controller('RecapSituationCtrl', function($scope, $s
...
+                if (! _.contains(ressourcesWithoutMonthlyAmount, ressourceType.id)) {
```
MattiSG: Searching for a given specific string in an array works as well with the native API, and will be faster and decrease your dependency on external libraries: [`ressourcesWithoutMonthlyAmount.indexOf(resourceType.id) < 0`](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Array/indexOf).

app/views/partials/foyer/recap-situation.html
fpagnoux: Only if you click on it.

app/js/controllers/foyer/recapSituation.js
fpagnoux: There is actually even simpler, thanks for putting me on the good path ;)

app/js/constants/ressources.js
```diff
@@ -201,7 +201,8 @@ angular.module('ddsCommon').constant('ressourceTypes', [
     {
         id: 'caMicroEntreprise',
         label: 'Micro-entreprise',
-        category: 'rpns'
+        category: 'rpns',
+        isMontantAnnuel: true
```
MattiSG: Huh?! This was supposed to already be in `app/constants/droits`!

MattiSG: Ok, never mind, ressources and droits are not the same thing  ;)

MattiSG:
GTM!

MattiSG:
Good job at refactoring, congrats  :)

PR #178, opened 2015-10-08
Use interpunct instead of standard point for gender uncertainty

fpagnoux:
Nouveau  :
image

Ancien :
image


MattiSG:
So awesome.

GTM

PR #181, opened 2015-10-09
Reorder situation recap (following family declaration reordering) 

fpagnoux:
Connected to #156 

Also fixes a bug introduced by #163


app/js/app.js
```diff
@@ -104,14 +104,14 @@ ddsApp.config(function($locationProvider, $stateProvider, $urlRouterProvider, $u
                 'individuForm@foyer.conjoint': individuFormView('conjoint')
             }
         })
-        .state('foyer.personnesACharge', {
-            url: '/personnes-a-charge',
+        .state('foyer.enfants', {
+            url: '/enfants',
```
MattiSG: :+1:

app/js/services/situationService.js
```diff
@@ -197,6 +197,32 @@ angular.module('ddsApp').factory('SituationService', function($http, $sessionSto
...
+
+        hasConjoint: function(situation) {
+            return _.find(situation.individus, { role: 'conjoint' });
+        },
+
+        setConjoint: function(situation, conjoint) {
+            var individus = situation.individus;
+            // si le conjoint existait déjà avant, on l'écrase
+            if (this.hasConjoint(situation)) {
+                individus[individus.length - 1] = conjoint;
```
MattiSG: This is probably not worth changing if it was already the case before and you just encapsulated the logic, but this is extremely fragile. It makes the implicit assumption that `situation.individus` is private and will only ever be changed through `setConjoint`, otherwise calling `setConjoint` would overwrite existing data.

app/js/services/situationService.js
```diff
@@ -197,6 +197,32 @@ angular.module('ddsApp').factory('SituationService', function($http, $sessionSto
...
+        },
+
+        setEnfants: function(situation, enfants) {
+            var individus = situation.individus;
+            individus = _.filter(individus, function(individu) {
+                return 'enfant' !== individu.role;
+            });
+            individus = individus.slice(0,1)
+                .concat(enfants)
+                .concat(individus.slice(1));
```
MattiSG: I think you want to [`splice`](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Array/splice) the array, really  :wink:

Beyond being shorter, understand that call to `slice` and `concat` creates a new array. This piece of logic is thus very costly.

test/spec/services/situationService.js
```diff
@@ -232,4 +232,77 @@ describe('Service: situationService', function () {
...
+                        { role: 'demandeur' },
+                        { role: 'enfant' }
+                    ]
+            };
+            var conjoint = { role : 'conjoint' };
+
+            // when
+            service.setConjoint(situation, conjoint);
+            // then
+            expect(situation.individus[2].role).toBe('conjoint');
```
MattiSG: Meh. Really, this is testing and rigidifying behaviour that's broken anyway. Is there any domain or architectural reason for storing the Conjoint at the end of all Individus? If there isn't any, it's just a matter of fragile, improper modelling. Testing this is just testing a private implementation detail.

Put otherwise: eating that the Conjoint is replaced if one exists, or that it is properly available once set (if possible, through getters) adds value. Testing that the Conjoint is the last element in an array doesn't add any, and actually decreases our ability to make architectural improvements later on.

MattiSG:
The complex array manipulation algorithm should be changed to `splice`. However, the tests and the fragile array implementation are fine for this PR to pass. We can improve it later on. It's already good that you improved coverage of the weak parts.

fpagnoux:
Ok, thanks for feedback, I'd be happy to discuss this in more details IRL or on slack :)

app/js/services/situationService.js
fpagnoux: With splice, the result would be : `[demandeur, [enfant1, enfant2], conjoint]` while we want `[demandeur, enfant1, enfant2, conjoint]`.
We could use `individus.splice.apply(...) but that would be less readable`

fpagnoux:
As we've seen, it may be good to have make `individus` an object with properties `demandeur`, `conjoint` and `enfants`. As we need to move forward with this PR, I won't do it here, but i'll keep that in mind.

PR #185, opened 2015-10-22
Ask taux d’invalidité question for children

fpagnoux:
Used to know if a handicapped adult should be considered à charge.

Connected to #169


MattiSG:
GTM

PR #190, opened 2015-10-30
Put step to any to stop wrong error messages

fpagnoux:
Connected to #179.


MattiSG:
GTM for your commit, but I added others, please review mine before merging.

PR #192, opened 2015-11-02
Disable pensions alimentaires browser validation

MattiSG:
Allow setting any value while keeping `step` amounts for arrows usage in input fields.
Connected to #179. Could supersede #190.


fpagnoux:
Oooook so there was `novalidate` on basically every form but this one ? 
It's working on chrome for me, GTM

MattiSG:
> Oooook so there was novalidate on basically every form but this one ?
Yup, hence the spotted inconsistency  ;)

PR #165, opened 2015-09-25
Conditions de régularité

fpagnoux:
Connected to #122

PR #195, opened 2015-11-06
Refactor individuform

fpagnoux:
Refactoring cherry-picking from #165

### DIFF
--- a/app/js/controllers/foyer/individuForm.js
+++ b/app/js/controllers/foyer/individuForm.js
@@ -53,7 +53,7 @@ angular.module('ddsApp').controller('FoyerIndividuFormCtrl', function($scope, in
     };
 
     var isIndividuParent = IndividuService.isRoleParent(individuRole);
-    $scope.individu = isIndividuParent && _.find($scope.situation.individus, { role: individuRole }) || DEFAULT_INDIVIDU;
+    $scope.individu = isIndividuParent && _.find($scope.situation.individus, { role: individuRole }) || _.cloneDeep(DEFAULT_INDIVIDU);
 
     $scope.individu.situationsPro.forEach(function(situationPro) {
         $scope.selectedStatuts[situationPro.situation] = true;
--- a/app/js/controllers/foyer/individuForm.js
+++ b/app/js/controllers/foyer/individuForm.js
@@ -3,12 +3,13 @@
 angular.module('ddsApp').controller('FoyerIndividuFormCtrl', function($scope, individuRole, situationsFamiliales, SituationService, IndividuService) {
     $scope.statutsSpecifiques = IndividuService.getStatutsSpecifiques();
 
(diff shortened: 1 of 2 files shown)


app/js/controllers/foyer/individuForm.js
```diff
@@ -3,22 +3,21 @@
 angular.module('ddsApp').controller('FoyerIndividuFormCtrl', function($scope, individuRole, situationsFamiliales, SituationService, IndividuService) {
     $scope.statutsSpecifiques = IndividuService.getStatutsSpecifiques();
 
-    var options = {};
-    options.captureRelationConjoint = (individuRole == 'conjoint');
-    options.checkNationalite = (individuRole == 'demandeur');
+    $scope.options = {};
+    $scope.options.captureRelationConjoint = (individuRole == 'conjoint');
+    $scope.options.checkNationalite = (individuRole == 'demandeur');
```
MattiSG: All this could be factored in a single object definition:

```js
$scope.options = {
    captureRelationConjoint = …,
    …
}
```

app/js/controllers/foyer/individuForm.js
```diff
@@ -48,23 +47,19 @@ angular.module('ddsApp').controller('FoyerIndividuFormCtrl', function($scope, in
...
+        autoEntrepreneurActiviteType: 'bic',
+        situationsPro: []
     };
 
-    if (true === ($scope.captureRelationConjoint = !! options.captureRelationConjoint)) {
-        $scope.individu.statutMarital = 'mariage';
-    }
-
     var isIndividuParent = IndividuService.isRoleParent(individuRole);
+    $scope.individu = isIndividuParent && _.find($scope.situation.individus, { role: individuRole }) || DEFAULT_INDIVIDU;
```
MattiSG: The line below means parts of the code could assign to `DEFAULT_INDIVIDU`!

To be safe, `_.cloneDeep(DEFAULT_INDIVIDU)`.

> The cleaner way to do it would be to deep [freeze](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Object/freeze) `Object.freeze(DEFAULT_INDIVIDU)` and, if in ES6, declare it with `const` and use [`Object.assign`](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Object/assign) rather than an `_`-provided equivalent.

MattiSG:
Everything that's written as a quote

> like this

is optional and mostly feedback.

app/js/controllers/foyer/individuForm.js
fpagnoux: Done

fpagnoux:
Thanks !
Is this GTM 
now ?

MattiSG:
GTM.

Seemingly depends on #202 too now…

PR #202, opened 2015-11-09
Minify images manually

MattiSG:
For no understandable reason, `imagemin` started failing all builds.
Anyway it's a good thing, compressing images on each build is pretty uselessly long.


fpagnoux:
Looks good to me

MattiSG:
Whoops no, missed removing the dependency from the `package.json`.

PR #212, opened 2015-11-24
Capture AEEH

fpagnoux:
Connected to #191


MattiSG:
Technically GTM, but please hold until discussion on UI capture is settled in #191.

MattiSG:
We'll take time to focus on giving specific widgets for each resource at some later point.

For the moment, let's just add a comment to remember we don't need to capture the actual amounts, and get this merged.

I'm handling it.

PR #223, opened 2015-12-09
Add deploy script

MattiSG:


deploy.sh
```diff
@@ -0,0 +1,67 @@
...
+
+git fetch origin $TARGET_BRANCH
+git checkout origin/$TARGET_BRANCH
+
+rm -rf node_modules # work around libsass bindings being never found
+npm install
+grunt build
+
+# Stop Mes Aides
+killall --user `whoami` node || echo 'No server was running'
```
fpagnoux: If we kill all node processes that belong to the user, I guess it means they can be only one instance per user.
So we create a new user each time we want to deploy a branch ?

deploy.sh
```diff
@@ -0,0 +1,67 @@
...
+set -ex
+
+# Defaults for production.
+# Example for development:
+#   PORT=8100 OPENFISCA_PORT=2000 OPENFISCA_CONFIG=development PUBLIC_URL=http://next.mes-aides.sgmap.fr ./deploy.sh aah
+TARGET_BRANCH=${1:-master}  # demo
+PORT=${PORT:-8000}  # 8100
+OPENFISCA_PORT=${OPENFISCA_PORT:-12000}  # 2000
+OPENFISCA_CONFIG=${OPENFISCA_CONFIG:-mes-aides}  # development
+PUBLIC_URL=${PUBLIC_URL:-https://mes-aides.gouv.fr}  # http://next.mes-aides.sgmap.fr
```
fpagnoux: So if I just pass another parameter, I can have, just using this script, an instance at any url .sgmap.fr I want ? Or do I need to create the url in another way, and then pass an appropriate parameter ?

MattiSG: No, this parameter is just here to be passed to the frontend. You have to set up the appropriate reverse proxy in NGINX, unfortunately  :(

deploy.sh
MattiSG: Good remark. With this implementation, indeed we would need to segregate by user.

To work around this, we can either use `pm2` or try to start the process through its full path and `killall` that path.

MattiSG: Do you want this to be explored in this PR or delay it until creating a user for each instance is too annoying? Even if we solved this problem here, there would still be some manual work to set up multiple instances, mostly in setting up the reverse proxy and creating a new OpenFisca ini file.

deploy.sh
```diff
@@ -0,0 +1,67 @@
...
+
+cd ..
+
+set +x
+
+echo "===================="
+echo "Déploiement effectué"
+echo "Attente du démarrage du serveur OpenFisca…"
+
+sleep 10
```
MattiSG: Rather use `wget --retry-connrefused --waitretry=1 --output-document=/dev/null http://localhost:$PORT:` if `wget` is installed by default.

deploy.sh
```diff
@@ -0,0 +1,67 @@
...
+if ! cd openfisca
+then
+    git clone https://github.com/sgmap/openfisca.git
+    cd openfisca
+fi
+
+git fetch origin $TARGET_BRANCH
+git checkout origin/$TARGET_BRANCH
+
+./update.sh --dev
```
fpagnoux: --dev is to avoid fetching and pull master, right ?
>The parameter name may be misleading, as we are not necessarily in dev mode. I had to go to check the `./update.sh` file to understand this intention.

deploy.sh
fpagnoux: Not necessarily, I guess it's not that hard to create one.

fpagnoux:
Seems GTM from the code, let me try it ;)

deploy.sh
MattiSG: Right, and remark is 100% apt too. I felt it when writing.

Do we change the default `update.sh` behaviour and make it stay on the current branch? I made it default to switch to `master` to ensure braindead calls would always _actually_ update to the latest stable version, not do nothing just because you were on a stale branch, but I end up almost always call it with `--dev`.

Options:
1. Always use current branch.
2. Rename `--dev` option to something like `--stay-on-branch` or `--current-branch`.
3. Require a target branch as a parameter (and allowing something like `--current-branch` as a valid target to avoid having to figure it out from git all the time).

deploy.sh
```diff
@@ -0,0 +1,67 @@
...
+git checkout origin/$TARGET_BRANCH
+
+rm -rf node_modules # work around libsass bindings being never found
+npm install
+grunt build
+
+# Stop Mes Aides
+killall --user `whoami` node || echo 'No server was running'
+# Start Mes Aides
+OPENFISCA_URL="http://localhost:$OPENFISCA_PORT" SESSION_SECRET=foobar NODE_ENV=production MES_AIDES_ROOT_URL="$PUBLIC_URL" PORT=$PORT MONGODB_URL="mongodb://localhost/$(whoami)" nohup node server.js >> ../mes-aides_log.txt &
```
fpagnoux: `MONGODB_URL="mongodb://localhost/$(whoami)"`
Does this mean one mongodb per user ? Would it be possible without too much work to always use the same mongo base ?

deploy.sh
fpagnoux: I'd go for 2, as this would be explicit *and* allow a simple use of `./update.sh` for developers not familiar with the project and git.

deploy.sh
MattiSG: That's one _database_, but a single Mongo server runs. It is better to keep them separate, you don't want to leak potentially broken stuff in the production database  ;)

deploy.sh
MattiSG: Implemented in `sgmap/openfisca#env`.

fpagnoux:
I tried it for ppa, seems GTM to me.

MattiSG:
Tested with `ppa` and `aah`. Waiting for `paris` to be available to confirm.

README.md
```diff
@@ -145,3 +143,33 @@ ssh deploy@sgmap.fr ./deploy
...
+
+```sh
+BRANCH=ppa
+adduser mes-aides-$BRANCH
+passwd -d mes-aides-$BRANCH
+mkdir /home/mes-aides-$BRANCH/.ssh/
+curl https://github.com/mesaides-bot.keys > /home/mes-aides-$BRANCH/.ssh/authorized_keys
+```
+
+#### Récupérer le script de déploiement
```
fpagnoux: Est-ce que cette série de commande doit aussi être effectuée en root, ou ça n'a pas d'importance ? 
C'est un peu ambigu tel que formulé aujourd'hui

MattiSG:
Instructions improved.

Tested with success with `paris`.

@fpagnoux is this GTM?

fpagnoux:
GTM

PR #240, opened 2016-01-28
Don't force rfr to zero

fpagnoux:
Connected to #198

### DIFF
--- a/app/js/controllers/foyer.js
+++ b/app/js/controllers/foyer.js
@@ -84,11 +84,7 @@ angular.module('ddsApp').controller('FoyerCtrl', function($scope, $state, $state
     });
 
     $scope.$on('rfr', function(e, rfr) {
-        if (rfr !== undefined) {
-            $scope.situation.rfr = rfr;
-            $scope.situation.ressourcesYearMoins2Captured = true;
-            $scope.$broadcast('ressourcesYearMoins2Captured');
-        }
+        $scope.situation.rfr = rfr;
         goToResultat();
     });
 
--- a/app/js/controllers/foyer.js
+++ b/app/js/controllers/foyer.js
@@ -84,9 +84,11 @@ angular.module('ddsApp').controller('FoyerCtrl', function($scope, $state, $state
     });
 
     $scope.$on('rfr', function(e, rfr) {
-        $scope.situation.rfr = rfr;
-        $scope.situation.ressourcesYearMoins2Captured = true;
-        $scope.$broadcast('ressourcesYearMoins2Captured');
+        if (rfr !== undefined) {
+            $scope.situation.rfr = rfr;
+            $scope.situation.ressourcesYearMoins2Captured = true;
+            $scope.$broadcast('ressourcesYearMoins2Captured');
+        }
         goToResultat();
     });
 
--- a/app/js/controllers/foyer/ressources/rfr.js
+++ b/app/js/controllers/foyer/ressources/rfr.js
@@ -7,6 +7,7 @@ angular.module('ddsApp').controller('FoyerRessourceRfrCtrl', function($scope) {
     $scope.rfr = $scope.situation.rfr;
 
     $scope.submit = function() {
+    	$scope.rfr = $scope.rfr !== undefined ? $scope.rfr : false; // rfr won't be updated in the database if the new value is undefined.
         $scope.$emit('rfr', $scope.rfr);
     };
 });
--- a/app/js/controllers/foyer/ressources/rfr.js
+++ b/app/js/controllers/foyer/ressources/rfr.js
@@ -1,9 +1,10 @@
+
 'use strict';
 
 angular.module('ddsApp').controller('FoyerRessourceRfrCtrl', function($scope) {
     $scope.yearMoins2 = moment($scope.situation.dateDeValeur).subtract('years', 2).format('YYYY');
     $scope.yearMoins1 = moment($scope.situation.dateDeValeur).subtract('years', 1).format('YYYY');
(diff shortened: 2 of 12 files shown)


app/js/controllers/foyer/ressources/rfr.js
```diff
@@ -3,7 +3,7 @@
 angular.module('ddsApp').controller('FoyerRessourceRfrCtrl', function($scope) {
     $scope.yearMoins2 = moment($scope.situation.dateDeValeur).subtract('years', 2).format('YYYY');
     $scope.yearMoins1 = moment($scope.situation.dateDeValeur).subtract('years', 1).format('YYYY');
-    $scope.rfr = $scope.situation.rfr || 0;
+    $scope.rfr = $scope.situation.rfr || undefined;
```
MattiSG: I don't think that's what you want to do. Setting a default value to `undefined` doesn't have much sense except preventing falsies.

In the case at hand: if `$scope.situation.rfr` is already `undefined`, this is useless. However, if the user explicitly entered `0`, you're ignoring the input (`0` is falsy, thus the value is the next operand).

app/js/controllers/foyer.js
```diff
@@ -84,9 +84,11 @@ angular.module('ddsApp').controller('FoyerCtrl', function($scope, $state, $state
     });
 
     $scope.$on('rfr', function(e, rfr) {
-        $scope.situation.rfr = rfr;
-        $scope.situation.ressourcesYearMoins2Captured = true;
-        $scope.$broadcast('ressourcesYearMoins2Captured');
+        if (rfr !== undefined) {
+            $scope.situation.rfr = rfr;
+            $scope.situation.ressourcesYearMoins2Captured = true;
```
MattiSG: This value is never reverted. What happens if the user goes back to the page and erases the fields?

fpagnoux:
Fixed.

app/js/services/situationService.js
```diff
@@ -226,6 +223,19 @@ angular.module('ddsApp').factory('SituationService', function($http, $sessionSto
                 .concat(enfants)
                 .concat(individus.slice(1));
             situation.individus = individus;
+        },
+
+        ressourcesYearMoins2Captured: function(situation) {
+            var ressourcesYM2Names = ['rncRevenusActivite', 'rncAutresRevenus', 'rncPensionsRetraitesRentes', 'fraisReelsDeductibles', 'rncPensionsAlimentaires', 'rncPensionsAlimentairesVersees'];
+            var ressourcesYM2 = [].concat.apply(
+                [],
```
MattiSG: More costly equivalent to `[].concat(situation…`.

app/js/services/situationService.js
```diff
@@ -226,6 +223,19 @@ angular.module('ddsApp').factory('SituationService', function($http, $sessionSto
...
+                situation.individus.map(function(individu) {
```
MattiSG: I don't get it: you concat an empty array with the result of a `map`, which is always an array? Why not simply take the return value immediately?

app/js/services/situationService.js
```diff
@@ -226,6 +223,19 @@ angular.module('ddsApp').factory('SituationService', function($http, $sessionSto
...
+                    return individu.ressources.filter(function(ressource) {
+                        return ressourcesYM2Names.indexOf(ressource.type) > 0;
+                    });
+                })
+            );
+            return ressourcesYM2.length > 0 || situation.rfr !== undefined;
```
MattiSG: Make the righthand condition a possible early exit at the beginning of the function to avoid a costly multiple iterations if `rfr !== undefined`.

app/js/services/situationService.js
```diff
@@ -226,6 +223,19 @@ angular.module('ddsApp').factory('SituationService', function($http, $sessionSto
...
                 .concat(individus.slice(1));
             situation.individus = individus;
+        },
+
+        ressourcesYearMoins2Captured: function(situation) {
+            var ressourcesYM2Names = ['rncRevenusActivite', 'rncAutresRevenus', 'rncPensionsRetraitesRentes', 'fraisReelsDeductibles', 'rncPensionsAlimentaires', 'rncPensionsAlimentairesVersees'];
+            var ressourcesYM2 = [].concat.apply(
+                [],
+                situation.individus.map(function(individu) {
+                    return individu.ressources.filter(function(ressource) {
```
MattiSG: > Use `some` rather than `filter` to avoid unnecessary calls  :)

app/js/services/situationService.js
```diff
@@ -226,6 +223,19 @@ angular.module('ddsApp').factory('SituationService', function($http, $sessionSto
                 .concat(enfants)
                 .concat(individus.slice(1));
             situation.individus = individus;
+        },
+
+        ressourcesYearMoins2Captured: function(situation) {
+            var ressourcesYM2Names = ['rncRevenusActivite', 'rncAutresRevenus', 'rncPensionsRetraitesRentes', 'fraisReelsDeductibles', 'rncPensionsAlimentaires', 'rncPensionsAlimentairesVersees'];
```
MattiSG: Constant, should be uppercase.

app/js/services/situationService.js
fpagnoux: To flatten the array of arrays returned by map. If I just return the array, the length will be the same as `individus` without checking whether they have N-2 ressources.
But I'm trying to find a more simple solution if this is too hard to grasp.

fpagnoux:
This is way simpler now, thanks for the feedback :+1:
However I still have a dependency on the api...

MattiSG:
GTM

PR #241, opened 2016-02-02
Use redux

MattiSG:
Base architecture on [Redux](http://redux.js.org) to build the OpenFisca situation gradually.
Make a generic handling of forms, with the assumption that each page presents one question that maps to one property of the OF situation.


front/actions.js
```diff
@@ -0,0 +1,39 @@
+export const UPDATE_OPENFISCA_SITUATION = 'UPDATE_OPENFISCA_SITUATION';
+export const FETCH_RESULTS = 'FETCH_RESULTS';
+export const GOT_RESULTS = 'GOT_RESULTS';
+export const ERROR = 'ERROR';
+
+
+export function updateOpenfiscaSituation(additional) {
```
MattiSG: Improper naming.

front/store.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+const store = createStore(storageMiddleware(reducer));
+
+export default store;
+
+export let subscribe = store.subscribe.bind(store);
+export let dispatch = store.dispatch.bind(store);
+export let getState = store.getState.bind(store);
```
MattiSG: Why not `const`? Why not export only `store`?

front/store.js
```diff
@@ -0,0 +1,90 @@
...
+                {
+                    error: action.error,
+                }
+            );
+        default:
+            return state;
+    }
+}
+
+export function storageMiddleware(reducer, storage) {
```
MattiSG: Should use the redux standard middleware facility, and possibly be in a separate file.

front/forms.js
```diff
@@ -0,0 +1,30 @@
...
+    const form = document.forms[0],
+          input = form.elements[inputName];  // not equivalent to querySelector due to radio buttons
+
+    form.addEventListener('submit', event => {
+        event.preventDefault();
+
+        const value = input.value;
+
+        if (input.required && ! value)
+            return store.dispatch(setError(name, 'required', value));
```
MattiSG: `name` doesn't exist, should be `inputName`.

webpack.config.js
```diff
@@ -0,0 +1,33 @@
...
+    },
+    output: {
+        path: './dist/js',
+        filename: '[name].js',
+    },
+    module: {
+        loaders: [
+            {
+                test: /store\.js$/,
+                loader: 'expose',
```
MattiSG: Leaking in the global scope is bad.

front/lib/redux-semantic-form.js
```diff
@@ -0,0 +1,11 @@
+store.subscribe(function updateErrorMessages() {
+    const state = store.getState();
+
+    Array.prototype.forEach.call(document.querySelectorAll('[data-for]'), message => {
+        message.classList[
+            message.attributes['data-for'].value == (state.error && state.error.id)
```
MattiSG: Unclear. Use an intermediary variable.

templates/layout.handlebars
```diff
@@ -17,5 +17,15 @@
     </head>
     <body>
         {{{ content }}}
+
+        {{#if isQuestion }}
```
MattiSG: Shouldn't this be another layout inheriting from the main layout?

webpack.config.js
MattiSG: Minimized exposure, still required for the temporary “recap” template. Will be removed once this template is properly coded.

templates/layout.handlebars
MattiSG: It should, but Hapi Handlebars doesn't support multiple layouts inclusion. This isn't so bad at the moment.

PR #252, opened 2016-03-11
Repair error display in ressource form

fpagnoux:
May be connected to #196.

Bug: When users leave a field blank, it is impossible to validate, but no error shows.

image

### DIFF
--- a/app/js/controllers/foyer/ressources/ressources.js
+++ b/app/js/controllers/foyer/ressources/ressources.js
@@ -93,9 +93,7 @@ angular.module('ddsApp').controller('FoyerRessourcesCtrl', function($scope, $sta
         };
     });
 
-    $scope.montantInvalide = function(montant) {
-        return ! angular.isNumber(montant);
-    };
+    $scope.isNumber = angular.isNumber;
 
     var applyIndividuVMRessourcesToIndividu = function(individuVM) {
         var individu = individuVM.individu;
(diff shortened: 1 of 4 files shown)


app/js/controllers/foyer/ressources/ressources.js
```diff
@@ -93,9 +93,7 @@ angular.module('ddsApp').controller('FoyerRessourcesCtrl', function($scope, $sta
         };
     });
 
-    $scope.montantInvalide = function(montant) {
```
fpagnoux: @MattiSG, 
I had trouble transmitting this function to this insulated scope of the directive, so as it was not doing much, I just got rid of the function.

app/js/directives/captureMontantRessource.js
```diff
@@ -11,13 +11,15 @@ angular.module('ddsApp').directive('montantRessource', function(SituationService
             ressourceType: '=',
             dateDeValeur: '=',
             index: '=',
-            onGoingLabel: '=?'
+            onGoingLabel: '=?',
+            form: '=',
```
fpagnoux: So that the directive knows if the form has been validated.

HelaG:
C'est ce qui explique les retours qu'on a eu "je n'arrive pas à obtenir un résultat" ?

fpagnoux:
Je pense oui.

app/js/controllers/foyer/ressources/ressources.js
MattiSG: :+1:

MattiSG:
GTM

PR #254, opened 2016-03-14
Paris refactor homepage

fpagnoux:
Suite au retour de d'Isabelle T. de Paris. 

Proposition : 
image


fpagnoux:
Une fois appuyé : image

fpagnoux:
Alternative sur un seul bloc: 
image

fpagnoux:
Paris est sur le bandeau principal : 

image

Ce n'est pas automatisé car je pense qu'aujourd'hui on a pas envie de mettre en valeur un partenaire sur lequel on a une seule aide "pilote", comme le 93.

fpagnoux:
@HelaG @MattiSG

app/js/services/resultatService.js
```diff
@@ -2,20 +2,25 @@
 
 angular.module('ddsApp').service('ResultatService', function($http, $modal, droitsDescription) {
 
+    var DROITS_POTENTIELS = droitsDescription.prestationsNationales.concat(
+        _.flatten(_.pluck(droitsDescription.partenairesLocaux, 'prestations'))
+    );
+
     // Si la valeur renvoyée par l'API vaut null, cela signifie par convention que l'aide a été injectée et non recaculée par le simulateur
     function sortDroits(droitsCalcules) {
+        console.log(droitsCalcules);
```
MattiSG: Committed `console.log`.

app/js/constants/droits.js
```diff
@@ -1,302 +1,322 @@
...
-        'isBaseRessourcesYearMoins2': true,
-        'uncomputability': {
-            'intervention_CDAPH_necessaire': {
-                'reason': 'le montant de l’AAH dépend du taux d’incapacité déterminé par la <abbr title="Commission des droits et de l’autonomie des personnes handicapées">CDAPH</abbr> après le dépôt de votre demande',
-                'solution': 'Contactez <a href="http://informations.handicap.fr/carte-france-mdph.php">la <abbr title="Maison départementale des personnes handicapées">MDPH</abbr> la plus proche</a> pour faire établir ce taux. Attention, votre éligibilité à l’AAH peut impacter vos droits au RSA et à la CMU-C.'
-            }
+    ],
+    'partenairesLocaux': [
+        {
+            'id': 'paris',
```
MattiSG: Why an array with `id` rather than an object?

app/js/services/resultatService.js
```diff
@@ -2,20 +2,25 @@
 
 angular.module('ddsApp').service('ResultatService', function($http, $modal, droitsDescription) {
 
+    var DROITS_POTENTIELS = droitsDescription.prestationsNationales.concat(
+        _.flatten(_.pluck(droitsDescription.partenairesLocaux, 'prestations'))
```
MattiSG: Pure JS equivalent:
```js
Array.prototype.concat.call(
    droitsDescription.prestationsNationales,
    droitsDescription.partenairesLocaux.map(function(partenaire) { return partenaire.prestations })
);
```

(no strong opinion here, I'm less familiar with `_` and `pluck` was not very explicit, but it is terser)

app/views/partials/homepage.html
```diff
@@ -83,4 +85,33 @@ <h2 class="text-center">Les aides simulées</h2>
...
+    <div ng-repeat="partenaireLocal in partenairesLocaux">
+      <hr>
+      <div class="row">
+        <div class="text-center col-md-2">
+          <img ng-src="/img/{{ partenaireLocal.imgSrc }}" alt=""/>
+          <h2>{{ partenaireLocal.label }}</h2>
+        </div>
+        <div class="col-md-10">
+          <div ng-repeat="droit in partenaireLocal.prestations">
+            <div class="clearfix" ng-if="showDetails && $index % 2 == 0"></div>
```
MattiSG: `$index % 2 == 0` ≣ `$even`

MattiSG: Why do you need a `clearfix` every two prestations? Looks like a hack for half-wide columns.

MattiSG:
Super ! Je préfère sur un seul bloc.

app/js/constants/droits.js
fpagnoux: I wanted to be able to choose the order and put Paris first.
For prestations, before they were objects with and id key, which was redundant.
Is it better to have an object ?

app/views/partials/homepage.html
fpagnoux: I adapted what was used for the above block. I will try to use a grid instead.

app/js/constants/droits.js
MattiSG: Well, I find it much easier to navigate a property path by names, like `partenairesLocaux.paris` rather than `partenairesLocaux[0]`, and to iterate over property names rather than indices…
The spec says the order is undefined, but most engines keep object inserting order. For such a non-critical presentation, I would say the risk of a different order on some browsers can be taken, but I don't care enough to ask for rewriting the navigations  ;)

app/views/partials/homepage.html
fpagnoux: It gets messy when the `clearfix`is removed. The top of the prestation cards are not aligned anymore, and it doesn't look great.

MattiSG: Ok, but what happens on a mobile device? I’m worried this solution isn’t responsive.

app/js/constants/droits.js
fpagnoux: Ok, I understand. I did have the wrong order on chrome though.
I would prefer not rewrite as I spent already too much time torturing myself with css ;)

app/js/services/resultatService.js
fpagnoux: You mean `Array.prototype.concat.apply` I guess :rabbit: 
I understand the idea, but it is not super transparent for me. Is `Array.prototype.concat.apply` the same than `[].concat.apply` ?

MattiSG: Yes, sorry  :wink:

> Is `Array.prototype.concat.apply` the same than `[].concat.apply` ?

Same effect, but not exactly the same thing happens. In the latter case, you instantiate a new `Array` to get access to its `concat` method, while in the former you just access to it straight to the already-instantiated `prototype` reference. Prefer the former for performance.

app/views/partials/homepage.html
fpagnoux: On the "mobile" mode of chrome dev tools, it just displays a single column of prestation instead of two per lines. I didn't try on a actual phone though. What is your process to check about responsiveness ?

MattiSG: Good. Then I’m fine with it.

> What is your process to check about responsiveness ?

I much prefer Safari’s [Responsive Design Mode](https://developer.apple.com/videos/play/wwdc2015/505/ <https://developer.apple.com/videos/play/wwdc2015/505/>), but both are fine to make a quick check when introducing design changes. Then, every once in a while (or for major changes) you make a smoke test with whichever devices are available around you. Just put them on the same local network as your development machine and access it through its IP  :)

fpagnoux: Ok, I doubled check with Safari. It is working well, except for this (that I already noticed with a friend's phone):

image

fpagnoux:
Is it now GTM ?

MattiSG:
GTM

PR #262, opened 2016-04-06
Use forever in prod

fpagnoux:
The goal is to make the feature branch deployment process more robust, so that we can use it in real prod as well.

To do:
- [x] Explicit dependency to forever in Readme
- [x] Disable openfisca live-reload in production mode
- [x] Explicit how to restart mongo if it crashes
- [ ] ~~Include to script prod specific nginx configuration (ssl, etc.)~~
- [ ] Make nginx serve a 404 for random addresses

Fixes sgmap/beta.gouv.fr#105.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,9 @@ Once you have Node and npm installed, run:
 npm install --global grunt-cli bower
 ```
 
+### In production
+
+[foreverjs](https://github.com/foreverjs/forever) is used to run the server in production. It needs to be installed before running the deployment script.
 
 Application
 -----------
@@ -147,6 +150,7 @@ ssh openfisca-mes-aides@sgmap.fr ./deploy
 ### Déployer une feature branch
 
 #### Ajouter un utilisateur capable de déployer sur le serveur de production
+> La taille du nom d'utilisateur étant limitée à 32 caractères sur le serveur de production, le nom de la feature branch ne doit pas dépasser 22 caractères.
 
 En se connectant en tant que `root`.
 
@@ -170,6 +174,6 @@ chmod u+x /home/mes-aides-$BRANCH/deploy.sh
 
 ```sh
 ssh-add ~/.ssh/mes-aides-bot
-ssh mes-aides-$BRANCH@sgmap.fr "PORT=8200 OPENFISCA_PORT=12200 ./deploy.sh $BRANCH"
+ssh mes-aides-$BRANCH@sgmap.fr "PORT=8200 OPENFISCA_PORT=12200 ./deploy.sh"
 ssh root@sgmap.fr "service nginx reload"
 ```
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ npm install --global grunt-cli bower
 
 [upstart](http://upstart.ubuntu.com/index.html) and [foreman](http://theforeman.org/) are used to run the production server for [mes-aides.gouv.fr](https://mes-aides.gouv.fr/).
 
-[foreverjs](https://github.com/foreverjs/forever) is used to run the server for staging feature branche versions of mes-aides. It needs to be installed before running the deployment scripts.
+[foreverjs](https://github.com/foreverjs/forever) is used to run the server for staging feature branches versions of mes-aides. It needs to be installed before running the deployment scripts.
 
 The production Mongo server can be (re)started with `ssh root@sgmap.fr "service mongod (re)start"`.
 
--- a/README.md
+++ b/README.md
@@ -181,3 +181,9 @@ ssh-add ~/.ssh/mes-aides-bot
 ssh mes-aides-$BRANCH@sgmap.fr "PORT=8200 OPENFISCA_PORT=12200 ./deploy.sh"
 ssh root@sgmap.fr "service nginx reload"
 ```
+
+### Supprimer une instance de feature branche
+```sh
+ssh mes-aides-$BRANCH@sgmap.fr 'forever stopall; rm /etc/nginx/conf.d/$(whoami).conf'
+ssh root@sgmap.fr "userdel mes-aides-$BRANCH"
+```
--- a/README.md
+++ b/README.md
@@ -192,7 +192,8 @@ ssh root@sgmap.fr "service nginx reload"
 ssh mes-aides-$BRANCH@sgmap.fr "./deploy.sh"
 ```
 
-### Supprimer une instance de feature branche
+### Supprimer une instance de feature branch
+
 ```sh
 ssh mes-aides-$BRANCH@sgmap.fr 'forever stopall; rm /etc/nginx/conf.d/$(whoami).conf'
 ssh root@sgmap.fr "userdel mes-aides-$BRANCH --remove"
--- a/README.md
+++ b/README.md
@@ -32,7 +32,11 @@ npm install --global grunt-cli bower
 
 ### In production
 
-[foreverjs](https://github.com/foreverjs/forever) is used to run the server in production. It needs to be installed before running the deployment script.
+[upstart](http://upstart.ubuntu.com/index.html) and [foreman](http://theforeman.org/) are used to run the production server for [mes-aides.gouv.fr](https://mes-aides.gouv.fr/).
+
+[foreverjs](https://github.com/foreverjs/forever) is used to run the server for staging feature branche versions of mes-aides. It needs to be installed before running the deployment scripts.
+
+The production Mongo server can be (re)started with `ssh root@sgmap.fr "service mongod (re)start"`.
 
 Application
 -----------
--- a/README.md
+++ b/README.md
@@ -153,8 +153,12 @@ ssh openfisca-mes-aides@sgmap.fr ./deploy
 
 ### Déployer une feature branch
 
-#### Ajouter un utilisateur capable de déployer sur le serveur de production
+Chaque feature branch est déployée sur le serveur de production par un utilisateur spécifique `mes-aides-$BRANCH`. Lorsque cet utilisateur exécute le script de déploiement `deploy.sh`, les branches `$BRANCH` de mes-aides-ui et [openfisca](https://github.com/sgmap/openfisca) sont déployées. Si cette branche n'existe pas sur le dépôt openfisca, la branche `master` est déployée.
+
 > La taille du nom d'utilisateur étant limitée à 32 caractères sur le serveur de production, le nom de la feature branch ne doit pas dépasser 22 caractères.
+
+
+#### Ajouter un utilisateur capable de déployer sur le serveur de production
 
 En se connectant en tant que `root`.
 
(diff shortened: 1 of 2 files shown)


deploy.sh
```diff
@@ -11,7 +11,8 @@ LOG_FILE=deployment.log
 #   PORT=8100 OPENFISCA_PORT=12100 ./deploy.sh aah
 # Example for production:
 #   PORT=8000 OPENFISCA_PORT=2000 PUBLIC_HOST=mes-aides.gouv.fr PROTOCOL=https ./deploy.sh
-TARGET_BRANCH=${1:-master}
+USER=`whoami`
+TARGET_BRANCH=${1:-${USER##mes-aides-}}
```
fpagnoux: This way the default branch will be the one corresponding to the user name, which avoids redundancy.

deploy.sh
```diff
@@ -40,34 +41,40 @@ npm install
...
 # Start Mes Aides
-OPENFISCA_URL="http://localhost:$OPENFISCA_PORT" SESSION_SECRET=foobar NODE_ENV=production MES_AIDES_ROOT_URL="$PROTOCOL://$PUBLIC_HOST" PORT=$PORT MONGODB_URL="mongodb://localhost/$(whoami)" nohup node server.js >> ../mes-aides.log 2>> ../mes-aides_error.log &
+OPENFISCA_URL="http://localhost:$OPENFISCA_PORT" SESSION_SECRET=foobar NODE_ENV=production MES_AIDES_ROOT_URL="$PROTOCOL://$PUBLIC_HOST" PORT=$PORT MONGODB_URL="mongodb://localhost/$USER" forever --uid mes-aides -l ../mes-aides.log -e ../mes-aides_error.log --append start server.js
 
 cd ..
 
 # Install OpenFisca
 if ! cd openfisca
 then
-    git clone https://github.com/sgmap/openfisca.git --branch $TARGET_BRANCH
```
fpagnoux: To be robust in case there is no `openfisca` target branch (which happens a lot in real life, when we are just trying to deploy something new on the front). Getting the master version is ok here, as this is just to get the `update.sh file`. The rest of the code will be updated by `update.sh` itself.

deploy.sh
```diff
@@ -40,34 +41,40 @@ npm install
...
+        echo "No branch $TARGET_BRANCH was found on sgmap/openfisca. Staying on current branch."
+        ./update.sh
+    }
 
 # Stop OpenFisca
-killall --user `whoami` /usr/bin/python || echo 'No OpenFisca server was running'
-# Start OpenFisca
-PORT=$OPENFISCA_PORT nohup ./start.sh mes-aides >> ../openfisca.log 2>> ../openfisca_error.log &
+{
+    forever stop openfisca
```
fpagnoux: @MattiSG, I could use your insights here.
It seems that, in this case, `forever stop` doesn't kill the `paster serve` process, but just stop running forever `start.sh`. And it seems that stopping to run start.sh doesn't stop `paster`.
I found two solutions: this one, with two commands, or not using `forever` here but in the `start.sh` script instead, when calling paster.

README.md
```diff
@@ -154,8 +154,7 @@ En se connectant en tant que `root`.
 BRANCH=ppa
 
 # Créer l'utilisateur
-adduser mes-aides-$BRANCH  # give whichever password and leave everything to default
-passwd --delete mes-aides-$BRANCH
+adduser mes-aides-$BRANCH --disabled-password --gecos ""
```
MattiSG: :+1: :+1: :+1: Thanks!!

deploy.sh
MattiSG: You probably don't need the double `##` (longest-match), a single one should suffice.

Do we still need to take `$1` as a parameter then? Sounds like a recipy for wasting an hour trying to understand that user `mes-aides-herp` actually deployed branch `derp`.

deploy.sh
```diff
@@ -40,34 +41,40 @@ npm install
...
 # Install OpenFisca
 if ! cd openfisca
 then
-    git clone https://github.com/sgmap/openfisca.git --branch $TARGET_BRANCH
+    git clone https://github.com/sgmap/openfisca.git
     cd openfisca
 fi
 
-./update.sh $TARGET_BRANCH
+./update.sh $TARGET_BRANCH || {
```
MattiSG: > The rest of this script is built with `if` constructs, it would probably be nicer to keep a uniform syntax:

```shell
if ! ./update.sh $TARGET_BRANCH
then
   …
fi
```

deploy.sh
MattiSG: Other solution: split [`start.sh`](https://github.com/sgmap/openfisca/blob/master/start.sh) functionality between creating the `current.ini` file and actually starting `paster`, and use `forever` here. You still need to keep `start.sh` in the other repo for easy usage, but it makes sense that in the current production context, we use another way to start the server.

> Using `forever` in `start.sh` is a no-go for me, as it adds complexity to the `openfisca-ops` module, which is supposed to be simple and is already quite complex as it handles workarounds. Even if it could make sense that it one day handles production constraints, I don't think `forever`, as a JS tool, makes sense as a dependency of a Python project.

MattiSG:
This is really clean bash code (except maybe for the `{ }` / `if` blocks inconsistency), congratulations! :+1:

deploy.sh
fpagnoux: The only time where I needed it was for a very long branch name, as there is a limit on the username size. So I had to use `mes-aides-irw` for `interrupted-ressources-wording, and actually use the extra argument.

deploy.sh
```diff
@@ -95,10 +108,17 @@ server {
...
         proxy_set_header Connection \"upgrade\";
 
-        proxy_pass  http://$(whoami);
+        proxy_pass  http://$USER;
         proxy_redirect off;
     }
-}" > /etc/nginx/conf.d/$(whoami).conf
+}" > /etc/nginx/conf.d/$USER.conf
+
+# Save current config
```
fpagnoux: So that, when an instance has been deployed once, we just need to run `./deploy.sh` without extra argument.
This would be necessary for prod: as we deploy it often, we don't want to have to specify the ports and public host each time.

README.md
```diff
@@ -171,6 +174,6 @@ chmod u+x /home/mes-aides-$BRANCH/deploy.sh
 
 ```sh
 ssh-add ~/.ssh/mes-aides-bot
-ssh mes-aides-$BRANCH@sgmap.fr "PORT=8200 OPENFISCA_PORT=12200 ./deploy.sh $BRANCH"
+ssh mes-aides-$BRANCH@sgmap.fr "PORT=8200 OPENFISCA_PORT=12200 ./deploy.sh"
 ssh root@sgmap.fr "service nginx reload"
```
fpagnoux: I ran into a case where `nginx reload` was not enough to actually take into account the new configuration. `nginx restart` was necessary.
Should we replace it here ? What makes me hesitate is:
- I'm not entirely sure this case would really happen in "real life", as it happened when I was trying things with feature branch deployment. I'm not sure either it can't happen.
- Is a `restart` more "risky" than a `reload` in term of prod ?

fpagnoux:
About mongo: I ran into this issue:
I was copying a database to another one, and the mongo server crashed. All the mes-aides instances were down. I managed to restart the server through `nohup mongod -f /etc/mongod.conf &`, and then redeployed one by one the instances.
I guess we should use forever for mongo too, right ?

README.md
MattiSG: poke @jdesboeufs

deploy.sh
```diff
@@ -95,10 +108,17 @@ server {
...
+echo "CURRENT_PORT=$PORT
+CURRENT_OPENFISCA_PORT=$OPENFISCA_PORT
+CURRENT_PUBLIC_HOST=$PUBLIC_HOST
+CURRENT_PROTOCOL=$PROTOCOL
+" > current_config
```
MattiSG: Add some extension (`.env`?) to make it clearer for exploratory.

deploy.sh
```diff
@@ -111,7 +131,7 @@ wget --quiet --retry-connrefused --waitretry=1 --output-document=/dev/null http:
 # Smoke test
 curl -sL -w "GET %{url_effective} -> %{http_code}\\n" localhost:$OPENFISCA_PORT -o /dev/null
 curl -sL -w "GET %{url_effective} -> %{http_code}\\n" localhost:$PORT -o /dev/null
-curl -sL -w "GET %{url_effective} -> %{http_code}\\n" $PROTOCOL://$PUBLIC_HOST -o /dev/null
+curl -sL -w "GET %{url_effective} -> %{http_code}\\n" $PROTOCOL://$PUBLIC_HOST/foyer/demandeur -o /dev/null
```
MattiSG: Does this fix anything? Adds friction to change.

deploy.sh
```diff
@@ -8,15 +8,22 @@ LOG_FILE=deployment.log
...
-PROTOCOL=${PROTOCOL:-http}
+USER=`whoami`
+TARGET_BRANCH=${USER#mes-aides-}
+source current_config || echo "No current_config file found."
+PORT=${PORT:-$CURRENT_PORT}
+OPENFISCA_PORT=${OPENFISCA_PORT:-$CURRENT_OPENFISCA_PORT}
 
+if ! [[ -n $PORT && -n $OPENFISCA_PORT ]]
+then
+    echo "Ports not specified, and not found in current_config file. Please provide them."; exit
```
MattiSG: `exit 1`, so it is known there was an error.

deploy.sh
fpagnoux: http://kjdqskjsdqkjfkjsdqkj.mes-aides.sgmap.fr/ -> 200

MattiSG: Yeah… Then nginx should serve a 404  :wink:

README.md
jdesboeufs: `nginx reload` => look at the configuration files and try to apply changes in _zero downtime_ way.
`nginx restart` => kill the process and restart nginx

You don't want to use `nginx restart` in your deployment ;)
If you just update or create a configuration file, there will be no problem.

fpagnoux:
It turns out the current production set up runs 8 mes-aides instances in parallel to improve performances.
Using only one with forever.js seems like a risky choice, especially as we expect more traffic in a near future, and as we today have no production metrics to anticipate GPU saturation.

I would suggest we merge this PR now (after reviewing), as it's already an improvement for feature branch deploying and documentation.

If we do want to go further in mastering our production environment, staying in node 0.10 seems a significant blocker (it prevents us to use pm2). It is not a priority for me right now, but hopefully we can invest some time on this in the next couple of months.

deploy.sh
```diff
@@ -8,15 +8,23 @@ LOG_FILE=deployment.log
...
-OPENFISCA_PORT=${OPENFISCA_PORT:-12000}
-PUBLIC_HOST=${PUBLIC_HOST:-$TARGET_BRANCH.mes-aides.sgmap.fr}
-PROTOCOL=${PROTOCOL:-http}
+USER=`whoami`
+TARGET_BRANCH=${USER#mes-aides-}
+source current_config.env || echo "No current_config.env file found."
+PORT=${PORT:-$CURRENT_PORT}
+OPENFISCA_PORT=${OPENFISCA_PORT:-$CURRENT_OPENFISCA_PORT}
 
+if ! [[ -n $PORT && -n $OPENFISCA_PORT ]]
```
MattiSG: Shouldn't this be `||` rather than `&&`? If any one of them isn't defined, we have a problem, right?

fpagnoux: Isn't it the current behaviour ?
NOT (A AND B) <=> (NOT A) OR (NOT B)

MattiSG: Damn you Morgan 😞 

> Didn't see that leading `!`  :wink:

MattiSG:
Great! GTM as is.
Any chance to include serving a 404 for random subdomains, as in the initial checklist, in this PR?

fpagnoux:
I don't think we can fix this problem with code in mes-aides-ui, it's probably some nginx and/or ovh configuration to do.

MattiSG:
GTM

PR #302, opened 2016-06-13
Make logement page accessible

fpagnoux:
Connected to #293 
Connected to #298

### DIFF
--- a/app/js/directives/yesNoQuestion.js
+++ b/app/js/directives/yesNoQuestion.js
@@ -3,17 +3,18 @@
 angular.module('ddsApp').directive('yesNoQuestion', function() {
     return {
         restrict: 'E',
+        transclude: true,
         templateUrl: 'partials/foyer/yes-no-question.html',
         scope: true,
         controller: 'yesNoQuestionCtrl',
         link: function ($scope, $element, $attributes) {
-            $scope.questionlabel = $attributes.questionlabel;
             $scope.modelString = $attributes.model;
         }
     };
 });
 
-angular.module('ddsApp').controller('yesNoQuestionCtrl', function($scope) {
+angular.module('ddsApp').controller('yesNoQuestionCtrl', function($scope, $transclude, $element) {
+    $element.find('legend').append($transclude())
 
 	// To get the object and the property described by the `model` directive attribute.
     function parse(modelString) {
--- a/app/js/directives/yesNoQuestion.js
+++ b/app/js/directives/yesNoQuestion.js
@@ -1,0 +1,33 @@
+'use strict';
+
+angular.module('ddsApp').directive('yesNoQuestion', function() {
+    return {
+        restrict: 'E',
+        templateUrl: 'partials/foyer/yes-no-question.html',
+        scope: true,
+        controller: 'yesNoQuestionCtrl',
+        link: function ($scope, $element, $attributes) {
+            $scope.questionlabel = $attributes.questionlabel;
+            $scope.modelString = $attributes.model;
+        }
+    };
+});
+
+angular.module('ddsApp').controller('yesNoQuestionCtrl', function($scope) {
+
+	// To get the object and the property described by the `model` directive attribute.
+    function parse(modelString) {
+        var re = /(.*)[.](.*)/;
+        var resultRegexp = re.exec(modelString);
+        return {
+            objectToModify: $scope.$eval(resultRegexp[1]),
+            propertyToModify: resultRegexp[2]
+        };
+    }
+
+    // ng-model doesn't accept a dynamic parameter, so we have to transmit the value manually.
+    $scope.updateModel = function() {
+    	var parsedModelString = parse($scope.modelString);
+        parsedModelString.objectToModify[parsedModelString.propertyToModify] = $scope.value;
+    };
+});
--- a/app/js/directives/yesNoQuestion.js
+++ b/app/js/directives/yesNoQuestion.js
@@ -13,12 +13,12 @@ angular.module('ddsApp').directive('yesNoQuestion', function() {
     };
 });
 
-angular.module('ddsApp').controller('yesNoQuestionCtrl', function($scope, $transclude, $element) {
+angular.module('ddsApp').controller('yesNoQuestionCtrl', function($scope, $transclude, $element, $interpolate) {
 	_.forEach($transclude(), function(elem) {
 		if (elem.localName == 'question') {
-			$element.find('legend').append(elem.innerHTML);
+			$element.find('legend').append($interpolate(elem.innerHTML)($scope));
 		} else if (elem.localName == 'help-block') {
-			$element.find('p').append(elem.innerHTML);
+			$element.find('p').append($interpolate(elem.innerHTML)($scope));
 		}
 	});
 
--- a/app/js/directives/yesNoQuestion.js
+++ b/app/js/directives/yesNoQuestion.js
@@ -8,27 +8,35 @@ angular.module('ddsApp').directive('yesNoQuestion', function() {
         scope: true,
         controller: 'yesNoQuestionCtrl',
         link: function ($scope, $element, $attributes) {
-            $scope.modelString = $attributes.model;
+            $scope.propertyPath = $attributes.model;
         }
     };
 });
 
 angular.module('ddsApp').controller('yesNoQuestionCtrl', function($scope, $transclude, $element) {
-    $element.find('legend').append($transclude())
+	_.forEach($transclude(), function(elem) {
+		if (elem.localName == 'question') {
+			$element.find('legend').append(elem.innerHTML);
+		} else if (elem.localName == 'help-block') {
+			$element.find('p').append(elem.innerHTML);
+		}
+	});
+
+    var parsedPropertyPath, targetObject;
 
 	// To get the object and the property described by the `model` directive attribute.
-    function parse(modelString) {
-        var re = /(.*)[.](.*)/;
-        var resultRegexp = re.exec(modelString);
+    function parse(propertyPath) {
+        var parts = propertyPath.split('.');
         return {
-            objectToModify: $scope.$eval(resultRegexp[1]),
-            propertyToModify: resultRegexp[2]
+            targetProperty: parts.pop(),
+            targetObject: parts.join('.')
         };
     }
 
     // ng-model doesn't accept a dynamic parameter, so we have to transmit the value manually.
-    $scope.updateModel = function() {
-    	var parsedModelString = parse($scope.modelString);
-        parsedModelString.objectToModify[parsedModelString.propertyToModify] = $scope.value;
+    $scope.updateValue = function() {
+        parsedPropertyPath = parsedPropertyPath || parse($scope.propertyPath);
+        targetObject = targetObject || $scope.$eval(parsedPropertyPath.targetObject);
+        targetObject[parsedPropertyPath.targetProperty] = $scope.value;
     };
 });
--- a/app/styles/front.scss
+++ b/app/styles/front.scss
@@ -698,7 +698,7 @@ legend {
   @extend .btn-default;
 }
 
-.invisibleInput {
+.screenReadersOnly {
   border: 0;
   clip: rect(0 0 0 0);
   height: 1px; margin: -1px;
--- a/app/styles/front.scss
+++ b/app/styles/front.scss
@@ -690,27 +690,8 @@ $autre-lighter: #cce4e7;
   cursor: help;
 }
 
-[data-toggle="buttons"] {
-  display: flex;
-  flex-direction: row;
-
-  button, label {
-    flex: 1;
-  }
-
-  .active .help-block, .active .help-col {
-    color: #D8D8D8;
-  }
-}
-
-@media (max-width: $screen-sm-min) {
-  [data-toggle="buttons"] {
-    flex-direction: column;
-  }
-}
-
 legend {
-  border-bottom: 0px;
+  border-bottom: none;
 }
 
 .radioButton {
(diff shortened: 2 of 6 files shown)


app/js/directives/yesNoQuestion.js
```diff
@@ -0,0 +1,40 @@
...
+        scope: true,
+        controller: 'yesNoQuestionCtrl',
+        link: function ($scope, $element, $attributes) {
+            $scope.modelString = $attributes.model;
+        }
+    };
+});
+
+angular.module('ddsApp').controller('yesNoQuestionCtrl', function($scope, $transclude, $element, $interpolate) {
+	_.forEach($transclude(), function(elem) {
```
fpagnoux: I'm kind of reimplementing pieces of angular here, but multi-slot transcluding is not implemented until 1.5, and without it we need to pass the texts in attributes, which is less convenient and have limitations.

app/js/directives/yesNoQuestion.js
```diff
@@ -0,0 +1,40 @@
...
+		if (elem.localName == 'question') {
+			$element.find('legend').append($interpolate(elem.innerHTML)($scope));
+		} else if (elem.localName == 'help-block') {
+			$element.find('p').append($interpolate(elem.innerHTML)($scope));
+		}
+	});
+
```
fpagnoux: The following is a little complex, but this is the best I found to be able to make this component able to take a `model` parameter. Any better implementation would be welcome, but I didn't find anything online.

fpagnoux:
I think I'm done for this page, @MattiSG what do you think ? 
There is two points I'm not sure about: 
- When the user chooses "Locataire" for instance, he.she will have a feedback (as it's a real radio input) and then will keep navigation through the form. Is this right, or do we need to do something to automatically put the focus on the next question ? 
- The help block is still after the radio boxes. This seems aligned with the idea that users just check it only when they are not sure about the question. Our testers seems surprised, but we stick to this choice, right ?

app/js/directives/yesNoQuestion.js
```diff
@@ -0,0 +1,40 @@
...
+	// To get the object and the property described by the `model` directive attribute.
+    function parse(modelString) {
+        var re = /(.*)[.](.*)/;
+        var resultRegexp = re.exec(modelString);
+        return {
+            objectToModify: $scope.$eval(resultRegexp[1]),
+            propertyToModify: resultRegexp[2]
```
MattiSG: Just `split` over `.`.

app/js/directives/yesNoQuestion.js
```diff
@@ -0,0 +1,40 @@
...
+			$element.find('p').append($interpolate(elem.innerHTML)($scope));
+		}
+	});
+
+	// To get the object and the property described by the `model` directive attribute.
+    function parse(modelString) {
+        var re = /(.*)[.](.*)/;
+        var resultRegexp = re.exec(modelString);
+        return {
+            objectToModify: $scope.$eval(resultRegexp[1]),
```
MattiSG: > IMO the `$scope.$eval` should be evaluated in a separate call, not in the parsing function.

app/styles/front.scss
```diff
@@ -314,6 +314,8 @@ $autre: #39747B;
 $autre-light: #a9d2d7;
 $autre-lighter: #cce4e7;
 
+
+/* Should be removed */
```
MattiSG: Under which circumstances? If it should be done now, do it now. If not, explain when it should be done.

app/styles/front.scss
```diff
@@ -688,21 +690,51 @@ $autre-lighter: #cce4e7;
...
-    flex: 1;
-  }
+.radioButton {
+  @extend .btn-default;
+}
 
-  .active .help-block, .active .help-col {
-    color: #D8D8D8;
-  }
+.invisibleInput {
```
MattiSG: Why `input`? Wouldn't this property set make anything invisible?

app/styles/front.scss
```diff
@@ -688,21 +690,51 @@ $autre-lighter: #cce4e7;
...
+  color: white;
+}
+
+input:checked + .radioButton > .help-col {
+  color: #D8D8D8;
+}
+
+.button-row {
+  display: flex;
+  flex-direction: row;
```
MattiSG: Have you done cross-browser testing for rendering?

app/views/partials/foyer/yes-no-question.html
```diff
@@ -0,0 +1,12 @@
+<fieldset class="form-group">
+  <legend class="col-sm-3 control-label"></legend>
+  <div class="col-sm-3">
+    <input type="radio" id="{{modelString}}-oui" name="{{ modelString }}" class="invisibleInput" ng-model="value" value="true" ng-change="updateModel()"/>
```
MattiSG: Template is named in English, ids are French-suffixed. Intentional?

app/js/directives/yesNoQuestion.js
```diff
@@ -0,0 +1,40 @@
...
+
+angular.module('ddsApp').directive('yesNoQuestion', function() {
+    return {
+        restrict: 'E',
+        transclude: true,
+        templateUrl: 'partials/foyer/yes-no-question.html',
+        scope: true,
+        controller: 'yesNoQuestionCtrl',
+        link: function ($scope, $element, $attributes) {
+            $scope.modelString = $attributes.model;
```
MattiSG: Should it really be called `ng-model` then, or is it misleading?

(I don't know exactly what `ng-model` entails, that's a real question)

app/js/directives/yesNoQuestion.js
fpagnoux: What if the argument is something like `situation.logement.loyer` ? If I use `split`, I then need to recombine the first 2 (or more) elements.

app/styles/front.scss
fpagnoux: 👍

app/views/partials/foyer/yes-no-question.html
fpagnoux: Nop.

app/js/directives/yesNoQuestion.js
MattiSG: Well, if nested property paths have to be supported, either the implementation is faulty or I don't understand why `$scope.$eval` wouldn't do all the work.

app/js/directives/yesNoQuestion.js
fpagnoux: As a developper seeing `ng-model`, I would think this is a built-in angular directive (and check the ng-model doc), and that would mislead me, so I'm more in favor of a custom keyword.

app/js/directives/yesNoQuestion.js
fpagnoux: The regexp is greedy, so it will stop at the last `.` it meets. The left part will be the object, and the right one the property we want to set.

MattiSG: Yes, I know. But the current implementation implies `$scope.$eval` is able to walk a property path at least until the last object in the chain. Why wouldn't it be able to walk it until the end then?

fpagnoux: Let's go back to the `situation.logement.loyer` example. If I'm using it, it means I have in my scope a `situation` object that contains a `logement` object.
So `$scope.$eval(situation.logement) = {...}`.
However, `$scope.$eval(situation.logement.loyer) = undefined`

If I want to set the property `loyer` of the already existing object `situation.logement`, I don't see another way than split the provided path at the last `.`.

MattiSG: Ok, gotcha. Sorry, I failed to take into account we were trying to set the value.

Implementation proposal with `split`:

```js
let parts = propertyPath.split('.');
return {
    targetProperty: parts.pop(),
    targetObject: parts.join('.')
}
```

fpagnoux: Thanks, that's much clearer 👍

MattiSG: Much cheaper too. Regexes are costly both to compile and execute  😃 

Beware of the behaviour if the given property path doesn't exist, though. You might be interested in checking it either [like this](https://github.com/MattiSG/Watai/blob/679619ed98a86ba52024960aa7816ba1cfe60e66/src/lib/mootools-additions.js#L4-L22) or through `$scope.$eval` (but does it throw if there's an undefined somewhere in the chain?

app/styles/front.scss
fpagnoux: IE 10 displays the blocks in a column but it is usable.

image

fpagnoux: IE11 is ok.
image

It should be fine for [other brothers](http://caniuse.com/#search=flex) and anyway, the use of `flex` does not change much with this PR.

fpagnoux:
Changes implemented, does this seem ok now ?

MattiSG:
> When the user chooses "Locataire" for instance, he.she will have a feedback (as it's a real radio input) and then will keep navigation through the form. Is this right, or do we need to do something to automatically put the focus on the next question ?
> The help block is still after the radio boxes. This seems aligned with the idea that users just check it only when they are not sure about the question. Our testers seems surprised, but we stick to this choice, right ?

MattiSG:
> do we need to do something to automatically put the focus on the next question ?

No, doing so would not be standard. As long as one can tab through the form, it's fine.

> The help block is still after the radio boxes.

Let's keep it for now. There was overuse of “speak what's after the button” due to broken tab navigation. If we still get bad feedback after fixing keyboard navigation we'll consider it.

app/js/directives/yesNoQuestion.js
```diff
@@ -0,0 +1,42 @@
...
+        return {
+            targetProperty: parts.pop(),
+            targetObject: parts.join('.')
+        };
+    }
+
+    // ng-model doesn't accept a dynamic parameter, so we have to transmit the value manually.
+    $scope.updateModel = function() {
+        parsedPropertyPath = parsedPropertyPath || parse($scope.propertyPath);
+        targetObject = targetObject || $scope.$eval(parsedPropertyPath.targetObject);
```
MattiSG: Is caching this useful? Is it not an error? When is `updateModel` called? We do want the `targetObject` to be updated if the property string changes.

app/styles/front.scss
```diff
@@ -688,21 +690,51 @@ $autre-lighter: #cce4e7;
   cursor: help;
 }
 
-[data-toggle="buttons"] {
-  display: flex;
-  flex-direction: row;
+legend {
+  border-bottom: 0px;
```
MattiSG: > If using the shorthand `border-<direction>` syntax, prefer using the `none` value rather than setting only the `border-width`; otherwise, it is expected that you're overriding just one property of the set, but you're currently disabling it altogether by setting its width to 0.

app/styles/front.scss
```diff
@@ -688,21 +690,51 @@ $autre-lighter: #cce4e7;
...
+  height: 1px; margin: -1px;
+  overflow: hidden;
+  padding: 0;
+  position: absolute;
+  width: 1px;
+}
+
+input:checked + .radioButton, input:checked + .radioButton:active:hover {
+  background-color: $btn-success-bg;
+  color: white;
```
MattiSG: > Risky: contrast (and thus readability) is defined by the difference between `background-color` and `color`, and only one of those is set through a variable. If `$btn-success-bg` is set to a light color, all these will become unreadable.

app/views/partials/foyer/logement.html
```diff
@@ -3,39 +3,32 @@
...
-        <button type="button" class="btn btn-sm btn-default" ng-model="logement.membreFamilleProprietaire" btn-radio="true">Oui</button>
-        <button type="button" class="btn btn-sm btn-default" ng-model="logement.membreFamilleProprietaire" btn-radio="false">Non</button>
-      </div>
-      <div class="col-sm-6">
-        <p class="help-block">Est-il un ascendant ou descendant de vous ou votre conjoint (fils, grand-père…)</p>
-      </div>
-    </div>
+    <yes-no-question model="logement.membreFamilleProprietaire" ng-if="captureMembreFamilleProprietaire()">
+      <question>Avez-vous un lien de parenté direct avec votre propriétaire&nbsp;?</question>
+      <help-block>Est-il un ascendant ou descendant de vous ou votre conjoint (fils, grand-père…)</help-block>
```
MattiSG: > Un exemple féminin ?  :wink:

app/js/directives/yesNoQuestion.js
fpagnoux: Err, you're right the function naming is misleading. It is called when the **value** of the field is changed, not the model.
The property string is not supposed to change for a given instance of this component.

If we assume the user directly chooses the right option, I agree there is not much utility, but it seemed awkward to potentially re-parse the path several times.

app/styles/front.scss
fpagnoux: Darn, `invisible` is a built-in bootstrap class with `visibility: hidden`. This breaks the accessibility mechanism (the form fields are ignored by the screen reader).

app/views/partials/foyer/logement.html
```diff
@@ -3,39 +3,32 @@
...
-        <button type="button" class="btn btn-sm btn-default" ng-model="logement.membreFamilleProprietaire" btn-radio="true">Oui</button>
-        <button type="button" class="btn btn-sm btn-default" ng-model="logement.membreFamilleProprietaire" btn-radio="false">Non</button>
-      </div>
-      <div class="col-sm-6">
-        <p class="help-block">Est-il un ascendant ou descendant de vous ou votre conjoint (fils, grand-père…)</p>
-      </div>
-    </div>
+    <yes-no-question model="logement.membreFamilleProprietaire" ng-if="captureMembreFamilleProprietaire()">
+      <question>Avez-vous un lien de parenté direct avec votre propriétaire&nbsp;?</question>
+      <help-block>Est-il un ascendant ou descendant de vous ou votre conjoint (enfant, grand-parent…)</help-block>
```
MattiSG: Missing punctuation.

MattiSG:
GTM

Testé au clavier sous Safari 9.1.1.

PR #303, opened 2016-06-16
Rendre accessible le choix des types de ressources

fpagnoux:
Connected to #300 

- Use native checkbox instead of buttons
- Hide icons from screen readers with aria-hidden ( **To check ?**)
- Display resources type in 3 columns
- Update angular to benefit from UI Bootstrap accessibility improvements (especially accordions).

---------
- [ ] Thorough cross-browser validation (as angular was updated).

### DIFF
--- a/app/styles/front.scss
+++ b/app/styles/front.scss
@@ -349,6 +349,8 @@ $autre-lighter: #cce4e7;
 
   .capture-rows {
     padding: 5px;
+    display: flex;
+    flex-wrap: wrap;
 
     .btn-ressources {
       @extend .btn-default;
@@ -357,6 +359,19 @@ $autre-lighter: #cce4e7;
     .btn-selection {
       font-size: 14px;
     }
+  }
+
+  .flexitem {
+    width: 33.33%;
+    text-align: center;
+  }
+
+  .flexitem > label {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 95%;
+    height: 90%;
   }
 }
 
--- a/app/styles/front.scss
+++ b/app/styles/front.scss
@@ -347,32 +347,7 @@ $autre-lighter: #cce4e7;
 
   }
 
-  .capture-rows {
-    padding: 5px;
-    display: flex;
-    flex-wrap: wrap;
 
-    .btn-ressources {
-      @extend .btn-default;
-    }
-
-    .btn-selection {
-      font-size: 14px;
-    }
-  }
-
-  .flexitem {
-    width: 33.33%;
-    text-align: center;
-  }
-
-  .flexitem > label {
-    display: flex;
-    align-items: center;
-    justify-content: center;
-    width: 95%;
-    height: 90%;
-  }
 }
 
 .capture-indemnites {
@@ -397,15 +372,6 @@ $autre-lighter: #cce4e7;
 
 .capture-rpns {
   @include ressource-category-capture($rpns, $rpns-light, $rpns-lighter);
-}
-
-.btn-selection {
-  // Reset border de 1px sur .btn
-  border-top: 0px;
-  border-bottom: 0px;
-  margin: 3px;
-  padding: 16px 16px;
-  font-size: 12px;
 }
 
 .intro {
@@ -707,10 +673,7 @@ $autre-lighter: #cce4e7;
 
 legend {
   border-bottom: none;
-}
-
-.radioButton {
-  @extend .btn-default;
+  font-size: inherit;
 }
 
 .screenReadersOnly {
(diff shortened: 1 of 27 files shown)


app/views/partials/droits-eligibles-list.html
```diff
@@ -1,10 +1,9 @@
-<accordion close-others="true">
-  <accordion-group is-open="droit.open"
-               class="resultats-accordion"
-               ng-repeat="(droitId, droit) in list"
-               ng-class="{'resultats-accordion-needs-n-2': droit.isBaseRessourcesYearMoins2 && !ressourcesYearMoins2Captured && ! isString(droit.montant), 'panel-opened': droit.open}">
+<uib-accordion close-others="true">
+  <uib-accordion-group is-open="droit.open"
+    class="resultats-accordion {{ getAccordionClass(droit) }}"
```
fpagnoux: `ng-class` and `uib-accordion` don't work well together, hence the hook with the extra function.

fpagnoux:
Before: 
image

After:
image

fpagnoux:
One drawback : 
On chrome on mac, one of the accessibility tags (`tabindex`) changes the focus behavior. This causes a blue outline around the header texts after the user click on them.

Removing it is tricky if we don't want to do it at the expense of accessibility.

image

In my opinion it is acceptable.

app/styles/front.scss
```diff
@@ -347,17 +347,7 @@ $autre-lighter: #cce4e7;
```
fpagnoux: Maybe this is more a question to discuss in real life, but what are the good practises in term of css files ? 
I feel rather lost when working with this unique big file, is there smart way to split the CSS ?

app/views/partials/droits-eligibles-list.html
MattiSG: :heart: for forward comment  :wink:

Might benefit from a comment though, if possible with a reference to the doc that explains why.

app/styles/front.scss
MattiSG: Yes, there are many. This is indeed extremely ugly.

The basic is to have a “global” / “main” CSS file, then additional ones for each screen.

When you have components though, it makes sense to have a file for each component.

We need a destination, and will refactor along along the way.

app/styles/front.scss
```diff
@@ -708,33 +686,60 @@ legend {
...
+
+    .inputButton {
+      display: flex;
+      align-items: center;
+      justify-content: center;
+      width: 100%;
+      height: 100%;
+      border-top: none;
+      border-bottom: none;
+      font-size: 14px;
```
MattiSG: Don't set the `font-size` in `px`. Make it a `%`, so that it varies accordingly to the parent font size.

app/styles/front.scss
```diff
@@ -708,33 +686,60 @@ legend {
...
+.button-grid {
+  display: flex;
+  flex-wrap: wrap;
 
-.button-row > .button-item > label {
-  width: 100%;
-  height: 100%;
+  .button-item {
+    width: 33.33%;
+    padding: 4px;
```
MattiSG: Don't set the padding in `px`. Set it in `em`, so that it varies accordingly to the inner font size.

app/views/partials/droits-eligibles-list.html
```diff
@@ -1,10 +1,9 @@
-<accordion close-others="true">
-  <accordion-group is-open="droit.open"
-               class="resultats-accordion"
-               ng-repeat="(droitId, droit) in list"
-               ng-class="{'resultats-accordion-needs-n-2': droit.isBaseRessourcesYearMoins2 && !ressourcesYearMoins2Captured && ! isString(droit.montant), 'panel-opened': droit.open}">
+<uib-accordion close-others="true">
```
MattiSG: Do we really want to `close-others`?

app/styles/front.scss
```diff
@@ -708,33 +686,60 @@ legend {
   width: 1px;
 }
 
-input:checked + .radioButton, input:checked + .radioButton:active:hover {
+.inputButton {
```
MattiSG: This name is not very clear to me… What's an “input button”?

MattiSG:
I feel pretty sweaty with the Angular upgrade. There were [many breaking changes](https://docs.angularjs.org/guide/migration#migrating-from-1-2-to-1-3), and I had many feedback that it was pretty hard to get all the edge cases right.

So, is the upgrade really necessary here? Can't we make the accordion accessible without upgrading?

If we do need the upgrade, we have to:

- [ ] Review all breaking changes from the [changelog](https://docs.angularjs.org/guide/migration#migrating-from-1-2-to-1-3) and check one by one whether they apply to our codebase.
- [ ] Do exhaustive cross-browser tests. That may justify adding basic integration tests.

app/views/partials/droits-eligibles-list.html
fpagnoux: Probably not.

app/styles/front.scss
fpagnoux: A button related to an input field (a checkbox or a radio), which appearance will change if the field is checked.
In other words the view of a checkbox or radio field under the shape of a button.

Can you think of another name ?

fpagnoux:
As `angular-bootstrap` rewrites the accordion, we cannot add the aria tags ourselves. There is no documented solution online, as this was a problem with pre-1.0 versions of angular bootstrap.

So if we want to keep the accordions and have them accessible, I do not see any other solution.

I already went through the changelog but I may have missed things. It seems like though we don't use many things in the wide angular world, so a lot doesn't concern us. We could peer on this, as re-reading it alone would bring anything new on my side.

fpagnoux:
Investing in integration testing would definitely mean we stick to this version though, let's make sure we discuss our refactoring strategy before doing it.

app/styles/front.scss
fpagnoux: @MattiSG

MattiSG:
Superseded by #311.

PR #309, opened 2016-06-29
Fix dependencies in recap embeded component

fpagnoux:
Bug introduced by #288 

`SituationService` depends on `ImpactStudyService` since this PR, and the module `ddsRecapSituation` depends on `SituationService`. So we need to pass `ddsRecapSituation` extra dependencies, even though it doesn't need it...

### DIFF
--- a/app/js/embed.js
+++ b/app/js/embed.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var app = angular.module('ddsRecapSituation', ['ui.router', 'ddsCommon']);
+var app = angular.module('ddsRecapSituation', ['ui.router', 'ddsCommon', 'angular-uuid']);
 
 app.config(function($locationProvider, $stateProvider) {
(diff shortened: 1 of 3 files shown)


fpagnoux:
@MattiSG Can you quickly review this or suggest a better fix ? ASI experts are going to add tests this afternoon and even though the bug it's not blocking, it isn't great.

app/js/embed.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-var app = angular.module('ddsRecapSituation', ['ui.router', 'ddsCommon']);
+var app = angular.module('ddsRecapSituation', ['ui.router', 'ddsCommon', 'angular-uuid']);
```
MattiSG: Why not add `angular-uuid` to `ddsCommon` directly?

fpagnoux: 👍

PR #311, opened 2016-06-29
Rendre accessible le choix des types de ressources

fpagnoux:
Connected to #300
Same as #303 with different target branch.

Should be merged when code is ok, integration testing will be done later.


app/styles/front.scss
```diff
@@ -708,33 +686,60 @@ legend {
   width: 1px;
 }
 
-input:checked + .radioButton, input:checked + .radioButton:active:hover {
+.inputButton {
```
MattiSG: Isn't this more like `input[type="radio"] + label, input[type="checkbox"] + label`?

MattiSG:
Closing this PR was a mistake, sorry. I don't know what happened. I probably clicked the button in a wrong tab this morning.

app/styles/front.scss
fpagnoux: Wouldn't that be too general ? What if somewhere there is a checkbox before a label where we don't want to apply this design ?

For instance the "Je continuerai à percevoir..."

MattiSG: Well, that's why I framed this as a question :wink:

The thing is, it's better to default to a uniform presentation, and that's what CSS is good at. Better have a specific opt-out class for the specific cases where you don't want to apply the style, than to repeat a specific class name in 90% of the cases. And I personally find it more expressive (clearer intention) to read “all labels that follow a radio button or a checkbox look like a button” than to read “the elements that have the `inputButton` class look like buttons”.

Now of course, it depends on how widely this behaviour is the expected one.

fpagnoux: image

fpagnoux: image

fpagnoux: Here are (almost) all the radios and checkboxes used in mes-aides. 

Some of them could be changed to buttons for more consistency, but there is still a bunch where a button would not be fitting.

MattiSG: Mmh. Seems to me you just outlined the parts of the interface we keep on thinking we should improve.
If you agree with this observation, then I'd much rather use this opportunity to single these bad parts out, so that we can get rid of both the specific CSS opt-outs and the additional classes once we've improved them, and keep better track of the work left to do.

fpagnoux: It seems that we can grab all the cases we want without a class-based selector (see last commit)

`input:checked + label` doesn't apply to the counter-examples I mentioned, as they (coincidently) have a structure like `<label><input></label>`.

Making the distinction based on this structure difference appears super dirty though...

MattiSG: Not so dirty to me. Having a difference in structure imply a styling difference sounds kind of normal. It's true that there is some ambiguity here, but it's not blocking to me.

MattiSG: As a side note, it's worth noting that if the `input` is not _in_ the `label`, then one needs to mark the relation up through `id` and `for` attributes… which is kind of heavy.

fpagnoux: 👍 
It actually caused trouble in another PR (to be reopened) as this id has to be unique in a page... I tried to put the the input inside the label (see last commit).

MattiSG:
Needs Angular update review.

fpagnoux:
Ok. To avoid being locked here, this PR was opened towards an integration branch, `access-integration`. If the code and UI (on our browsers) is ok, we should merge this so that I can open the depending PR. 

Is this GTM from in the integration branch ?

PR #312, opened 2016-06-29
Split family resources declaration

fpagnoux:
Depends on #311 
Connected to #301


MattiSG:
It is too hard to review a large PR that depends on another large changeset. Let's reopen this one when #311 is merged.

PR #315, opened 2016-06-30
Add continuous deployment

MattiSG:
Tested on several commits, including [build 2300](https://travis-ci.org/sgmap/mes-aides-ui/builds/141329936). This build was used to restart the service after a crash and proves starting works.


fpagnoux:
Nice, GTM 👍

PR #319, opened 2016-07-15
Deprecate openfisca ops

fpagnoux:
Goal and approach
-------------------

This PR aims at showing how we could use the benefits of the now strict versioning of openfisca to deprecate the meta-repo github.com/sgmap/openfisca, to open a discussion.

The deployment process we use for feature branch has been therefore rewritten without using this meta-repo.

Feasibility
----------
The dependencies towards `core` and `parsers` are now managed through versioning, so we don't need to worry about them.

The dependency between `france` and `web-api` is weak: `france` just needs to expose its TaxBenefitSystem so that the api can serve it. I don't see much risk of incompatibility.

However, we want to protect ourselves from breaking changes in `france` (e.g. renaming) and in `web-api` (e.g. route changes). **So we would have to store somewhere (openfisca_config.ini ?) the  major versions we want to install for the two packets.** Or the exact version, if we are afraid of the risk of undetected breaking changes.

The extensions are a little more annoying, as we have less control over their repository, and so far they are not released, versioned, or checked with CI. To be safe, we should thus **store the sha1 of the commit** we want to deploy, and run all tests when our partners want to update the code.

Strategy
--------

The fact that we still have to save somewhere 2 version numbers and a sha1 (and more if others extensions are coming) make me question the interest of deprecating the repository. A different strategy could be to simplify the meta repo by just removing references to `core` and `parsers`.

I still think deprecating it would have some advantages, as all what is necessary to run mes-aides would be on only one repository. Also, as the versioning is more clean now, it is less likely that other openfisca users use the meta-repository.

@MattiSG , what are your thoughts ?

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -65,7 +65,7 @@ fi
 
 pip install --requirement mes-aides-ui/openfisca/requirements.txt $user_option
 
-sed s/%PORT%/$OPENFISCA_PORT/ mes-aides-ui/openfisca/api_config.ini > current_openfisca_config.ini
+sed s/"port = 2000"/"port = $OPENFISCA_PORT"/ mes-aides-ui/openfisca/api_config.ini > current_openfisca_config.ini
 
 # Stop OpenFisca
 forever stop openfisca || echo 'No OpenFisca server was running'
--- a/deploy.sh
+++ b/deploy.sh
@@ -34,11 +34,11 @@ echo "OpenFisca expected to run on $OPENFISCA_PORT" >> $LOG_FILE
 echo "Server expected to be visible on $PUBLIC_HOST" >> $LOG_FILE
 
 # Install Mes Aides
-if ! cd mes-aides-ui
-then
+if [ ! -d "mes-aides-ui" ]; then
     git clone https://github.com/sgmap/mes-aides-ui.git --branch $TARGET_BRANCH
-    cd mes-aides-ui
 fi
+
+cd mes-aides-ui
 
 git fetch origin $TARGET_BRANCH
 git checkout origin/$TARGET_BRANCH
@@ -58,26 +58,20 @@ OPENFISCA_URL="http://localhost:$OPENFISCA_PORT" SESSION_SECRET=foobar NODE_ENV=
 cd ..
 
 # Install OpenFisca
-if ! cd openfisca
-then
-    git clone https://github.com/sgmap/openfisca.git
-    cd openfisca
+
+if [[ ! $VIRTUAL_ENV ]]; then
+    user_option='--user'
 fi
 
-if ! ./update.sh $TARGET_BRANCH
-then
-    echo "No branch $TARGET_BRANCH was found on sgmap/openfisca. Staying on current branch."
-    ./update.sh
-fi
+pip install --requirement mes-aides-ui/openfisca/requirements.txt $user_option
 
-PORT=$OPENFISCA_PORT ./generateMesAidesConfig.sh
+sed s/%PORT%/$OPENFISCA_PORT/ mes-aides-ui/openfisca/api_config.ini > current_openfisca_config.ini
 
 # Stop OpenFisca
 forever stop openfisca || echo 'No OpenFisca server was running'
 
 # Start OpenFisca
-forever --uid openfisca -l ../openfisca.log -e ../openfisca_error.log --append start -c "paster serve" config/current.ini
-cd ..
+forever --uid openfisca -l openfisca.log -e openfisca_error.log --append start -c "paster serve" current_openfisca_config.ini
 
 # Set up reverse proxy
 
--- a/openfisca/api_config.ini
+++ b/openfisca/api_config.ini
@@ -1,0 +1,49 @@
+# OpenFisca-Web-API - Development environment configuration
+#
+# The %(here)s variable will be replaced with the parent directory of this file.
+
+[DEFAULT]
+debug = true
+
+[server:main]
+use = egg:Paste#http
+host = 0.0.0.0
+port = %PORT%
+
+[app:main]
+use = egg:OpenFisca-Web-API
+country_package = openfisca_france
+log_level = DEBUG
+reforms =
+    openfisca_france.reforms.aides_cd93.aides_cd93
+extensions =
+    openfisca_paris
+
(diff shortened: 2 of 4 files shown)


deploy.sh
```diff
@@ -55,29 +55,52 @@ mongo localhost --eval "db.copyDatabase('mes-aides-master', '$USER')"
 # Start Mes Aides
 OPENFISCA_URL="http://localhost:$OPENFISCA_PORT" SESSION_SECRET=foobar NODE_ENV=production MES_AIDES_ROOT_URL="$PROTOCOL://$PUBLIC_HOST" PORT=$PORT MONGODB_URL="mongodb://localhost/$USER" forever --uid mes-aides -l ../mes-aides.log -e ../mes-aides_error.log --append start server.js
 
-cd ..
+cd ~
```
MattiSG: This enforces user scope rather than directory scope. Why?

openfisca/api_config.ini
```diff
@@ -0,0 +1,52 @@
...
+#
+# The %(here)s variable will be replaced with the parent directory of this file.
+
+[DEFAULT]
+debug = true
+# Uncomment and replace with the address which should receive any error reports
+email_to = mes-aides@sgmap.fr
+smtp_server = localhost
+error_email_from = openfisca-mes-aides@sgmap.fr
+error_from = openfisca-mes-aides@sgmap.fr
```
MattiSG: Outdated.

openfisca/api_config.ini
```diff
@@ -0,0 +1,52 @@
...
+use = egg:Paste#http
+host = 0.0.0.0
+port = %PORT%
+
+[app:main]
+use = egg:OpenFisca-Web-API
+country_package = openfisca_france
+log_level = DEBUG
+reforms =
+	openfisca_france.reforms.aides_cd93.aides_cd93
```
MattiSG: Missing Paris?

openfisca/versions.env
```diff
@@ -0,0 +1,3 @@
+OPENFISCA_API_VERSION=1.1.0.dev0
+OPENFISCA_FRANCE_VERSION=4.0.5
+OPENFISCA_PARIS_VERSION=3ca5d0e9f27484cbce1795527e0a0a4c63085fcc
```
MattiSG: Why not simply make all those a `requirements.txt` and `pip install`?

MattiSG:
I'm fine with deprecating `ops` if OpenFisca cross-components dependencies can now be installed easily.

The current suggested changeset doesn't convince me it would be much easier without, though  :wink:

Why can't we simply use `pip` to install the dependencies by simply specifying the target packages + versions / repos + SHAs targets?

MattiSG:
This also raises the question of how to properly address increasing extensions if external providers come in. That can probably wait until we have more than 10, though  ;)

openfisca/api_config.ini
fpagnoux: Paris is not a reform anymore, it's an extension managed by adding the directory in the right place. See [here](https://github.com/sgmap/mes-aides-ui/pull/319/files#diff-60254338249f657a0a83f98258a56bfeR85).

 Ideally we would also do the same for CD93, but as it's not moving much, i don't think it's worth investing time on it.

fpagnoux:
Here is a new proposition, based on a `requirements.txt` file, and some work on the openfisca side.

To deploy a feature branch that needs a specific openfisca feature branch, the developper has to edit this file.

We lose the automatic synchronisation based on naming between mes-aides and openfisca branches, but it seems reasonable to me.

I tested it for this branch and

openfisca/api_config.ini
```diff
@@ -0,0 +1,49 @@
...
+#
+# The %(here)s variable will be replaced with the parent directory of this file.
+
+[DEFAULT]
+debug = true
+
+[server:main]
+use = egg:Paste#http
+host = 0.0.0.0
+port = %PORT%
```
MattiSG: > If we're not being generic, we could as well use a default here, and replace the whole line through `sed`.

openfisca/api_config.ini
```diff
@@ -0,0 +1,49 @@
+# OpenFisca-Web-API - Development environment configuration
+#
+# The %(here)s variable will be replaced with the parent directory of this file.
+
+[DEFAULT]
+debug = true
```
MattiSG: > Do we really want this in production?

MattiSG:
I think making the target references explicit in the `requirements` is actually better than convention names.

I'm ok for trying this with our new partner on Friday. I'll leave it up to you to decide whether we should merge this now and possibly update (or even revert) it afterwards, or if it's better to wait until field-tested.

openfisca/api_config.ini
```diff
@@ -0,0 +1,49 @@
...
+port = %PORT%
+
+[app:main]
+use = egg:OpenFisca-Web-API
+country_package = openfisca_france
+log_level = DEBUG
+reforms =
+    openfisca_france.reforms.aides_cd93.aides_cd93
+extensions=
+    openfisca_paris
```
MattiSG: Why is one a `reform` and the other an `extension`?

fpagnoux: Only historical reasons:
- At first, only `reforms` were available. The API was pretty awful but we dealt with it to build POCS with Paris and CD93.
- When we started to work more seriously with Paris, we built the extension mechanism and got rid of the Paris reform
- We never paid the relatively small cost of turing `cd93` into an extension since as of today it a dead project.

Do you think consistency is worth paying the additional cost (I guess 1-3 hours) ?

openfisca/api_config.ini
fpagnoux: If set to `false`, openfisca will try to send emails when an error happens. We never used it before, we could try, but there may be heavy configuration to be able to send emails.

openfisca/api_config.ini
fpagnoux: 👍

openfisca/api_config.ini
MattiSG: Probably not by itself. It’ll probably be more surprising for our upcoming new contributors, but we’re still hoping to get the CD93 to commit some more aides, so let’s assign that task to them  ;)

PR #320, opened 2016-07-21
Add integration tests

fpagnoux:
PR opened for discussion, probably not ready to be merged as is.

Two suites of scenarios have been tested, to check the ability of having several suites.

IMO we now need to:

- Get some feedback on the already written tests.
- Discuss what level of test coverage we will consider sufficient to validate the angular upgrade.
- Run the tests on other browsers than chrome.

@MattiSG @Flightan


app/views/partials/foyer/capture-montant-ressource.html
```diff
@@ -8,14 +8,14 @@ <h3 class="panel-title">
...
-        <label for="{{ index + ressource.type.id + month.id }}">
+        <label for="{{ ressource.type.id }}_month{{ $index + 1 }}_{{ index }}">
           {{ month.label|uppercaseFirst }}
         </label>
         <span class="input-group">
           <input
             type="number"
             class="form-control text-right"
-            id="{{ index + ressource.type.id + month.id }}"
+            id="{{ ressource.type.id }}_month{{ $index + 1 }}_{{ index }}"
```
fpagnoux: This `id`, which started with a number, had to be changed to be selected.

app/views/partials/foyer/recap-situation.html
```diff
@@ -3,7 +3,7 @@ <h3 class="text-center">Votre situation</h3>
 <h4 class="section-header">Votre foyer</h4>
 <individu-block ng-repeat="individu in situation.individus" individu="individu"></individu-block>
 
-<div ng-if="situation.logement">
+<div id="logement-block" ng-if="situation.logement">
```
fpagnoux: We minimisez the amounts of added `id` to the CSS, but the recap is hard to navigate as is.

test/integration/components/RecapComponent.js
```diff
@@ -0,0 +1,8 @@
+firstPersonBirthdate: 'individu-block:nth-of-type(1) strong',
+secondPersonBirthdate: 'individu-block:nth-of-type(2) strong',
+thirdPersonBirthdate: 'individu-block:nth-of-type(3) strong',
+fourthPersonBirthdate: 'individu-block:nth-of-type(4) strong',
```
fpagnoux: This seems repetitive. Is there a way to factorise it ?

app/views/partials/foyer/capture-montant-ressource.html
MattiSG: Huh? Weird. That shouldn't be a problem:

> the id attribute value must be unique
> There are no other restrictions on what form an ID can take; in particular, IDs can consist of just digits, **start with a digit**, start with an underscore, consist of just punctuation, etc.

[Spec](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute)

MattiSG: No need to investigate much more though, the new name scheme is clearly better.

app/views/partials/foyer/recap-situation.html
MattiSG: > Avoid suffices like `block` that don't carry much meaning. We are always targeting DOM elements, so it's quite expected to be a container. `#logement` is probably clear enough. If you want to be more explicit and avoid ambiguity between inputs and recap, use BEM-like naming such as `recap-logement`.

test/integration/base-suite/3 - Add no child scenario.js
```diff
@@ -0,0 +1,8 @@
+description: 'Add no child.',
+
+steps: [
+    {
+        'EnfantsFormComponent.title': 'Les enfants de votre foyer',
```
MattiSG: > I'd recommend to use larger-encompassing RegExps such as `/enfants/` to increase resilience towards phrasing changes.

test/integration/base-suite/6 - Declare salary scenario.js
```diff
@@ -0,0 +1,18 @@
+description: 'Declare salary.',
+
+steps: [
+    ResourcesComponent.declareResources(),
+    ResourcesComponent.declareRevenuActivite(),
+    ResourcesComponent.salary(),
```
MattiSG: Missing a verb.

test/integration/base-suite/6 - Declare salary scenario.js
```diff
@@ -0,0 +1,18 @@
...
+    ResourcesComponent.submit(),
+    ResourcesComponent.setFirstMonthInput(1000),
+    ResourcesComponent.setSecondMonthInput(1000),
+    ResourcesComponent.setThirdMonthInput(1000),
+    ResourcesComponent.setLast12MonthsInput(12000),
+    ResourcesComponent.submit(),
+    RecapComponent.salaryTab(),
```
MattiSG: Missing a verb.

test/integration/components/RecapComponent.js
MattiSG: The best you can do is to declare a helper function in a `Fixture` file and call it:

```js
// IndividuFixture
function nthIndividu(index) {
     return 'individu-block:nth-of-type(' + index + ') strong';
}

// RecapComponent
firstPersonBirthdate: nthIndividu(1),
secondPersonBirthdate: nthIndividu(2),
…
```

test/integration/components/RecapComponent.js
```diff
@@ -0,0 +1,8 @@
...
+housingStatus: '#logement-block b',
+salaryTabLink: '.recap-revenusActivite a',
+salaryFirstMonth: '.recap-revenusActivite .row :nth-child(2)',
+salaryLast12Months: '.recap-revenusActivite hr + .row :nth-child(2)',
```
MattiSG: > For increased readability, keep spaces around selectors, not pseudo-classes (the ones after a `:`), for consistency with actual combined classes selectors.

test/integration/run-integration-tests.sh
```diff
@@ -0,0 +1,3 @@
+for element in *-suite
+   do watai "$element"
+done
```
MattiSG: You need to exit with a nonzero exit code if any of the execution returns a nonzero exit code.

test/integration/components/RecapComponent.js
fpagnoux: Not sure I understand. `.recap-revenusActivite hr + .row:nth-child(2)` would have a different behaviour. Can you explicit the change you recommend ?

MattiSG:
Very good!

For running tests with other browsers, iterate over browser combinations in the shell script:

```
--config '{"driverCapabilities":{"browserName":"'"$browser_name"'","version":"'"$browser_version"'"}}'
```

IMO a full end-to-end test like this is quite good to start with. I'd be more at ease if we also visited n-2 and one aide details screen.

test/integration/components/RecapComponent.js
fpagnoux: @MattiSG

MattiSG: I'm stupid. Forget it.

test/integration/components/RecapComponent.js
fpagnoux: Not sure the little improvement of readability in the component is worth the extra file.

So I guess there is no way to have dynamic name for components elements, like :


```js
// Scenario
nthPersonBirthdate(1): DEMANDEUR_BIRTHDATE
nthPersonBirthdate(2): CONJOINT_BIRTHDATE
nthPersonBirthdate(3): CHILD_BIRTHDATE

// Component
nthPersonBirthdate: function(index) {
     return 'individu-block:nth-of-type(' + index + ') strong';
}
```

test/integration/run-integration-tests.sh
fpagnoux: Ok, I added `set -e`

test/integration/components/RecapComponent.js
MattiSG: Nope…

MattiSG:
Isn't `watai` missing from the `devDependencies`?

MattiSG:
Running in Firefox clearly shows issues with the broken scroll on screen change…

fpagnoux:
To me the tests are ready, but I was not able to run them through saucelabs.

@MattiSG, I'll let you check the tests are runnable with other browsers ;)

fpagnoux:
To run tests in Saucelabs: 

- Before running tests, set the following environment variables

> This is necessary to connect to the distant selenium server through watai:

```sh
export SAUCE_USERNAME=(...)
export SAUCE_ACCESS_KEY=(...)
```
- Run sauce connect
```sh
sc --user $SAUCE_USERNAME --api-key $SAUCE_ACCESS_KEY
```
- Adapt the watai config file:

> Setting the platform is necessary for testing on IE, otherwise sauce labs will use windows XP.

```json
    "seleniumServerURL": {
        "hostname" : "ondemand.saucelabs.com",
        "port" : 80
    },
    "driverCapabilities": {
        "platform" : "Windows 7"
    }
```

fpagnoux:
Report on test run so far: 

https://framacalc.org/mes-aides-test-integration.

MattiSG:
Ok, passes under: Chrome 51, Chrome 48, Firefox 45, IE 10, IE 11, Firefox 45.

fpagnoux:
Is it good to rebase and merge then ?

app/styles/front.scss
```diff
@@ -637,10 +637,7 @@ $autre-lighter: #cce4e7;
     color: #888;
   }
 
-  .montant::after {
-    @extend .fa;
-    @extend .text-warning;
-    content: $fa-var-warning;
+  i {
```
MattiSG: This is a much, much larger selector. And I find it weird that we have to set these properties ourselves, it looks like basic positioning that should be provided by Bootstrap. Can't we get rid of the block altogether?

app/views/partials/foyer/capture-montant-ressource.html
```diff
@@ -29,14 +29,14 @@ <h3 class="panel-title">
...
-        <label for="{{ index + ressource.type.id }}-annuel">
+        <label for="{{ ressource.type.id }}_annuel_{{ index }}">
           Total depuis {{ debutAnneeGlissante }}
         </label>
         <span class="input-group">
           <input
             type="number"
             class="form-control text-right"
-            id="{{ index + ressource.type.id }}-annuel"
+            id="{{ ressource.type.id }}_annuel_{{ index }}"
```
MattiSG: Language and word inconsistency: `month` above, `annuel` here.

Maybe `{{ ressource.type.id }}-year_{{ index }}-month_{{ $index + 1 }}` and `{{ ressource.type.id }}-year_{{ index }}`?

app/views/partials/foyer/ressources/types.html
```diff
@@ -25,11 +25,10 @@
               ng-repeat="ressourceType in ressourceTypesByCategories[category.id]| orderBy:['positionInList','label']">
               <input
                 type="checkbox"
-                id="{{ ressourceType.id }}"
                 class="screenReadersOnly"
                 ng-model="individuVM.selectedRessourceTypes[ressourceType.id]"
               />
-              <div for="{{ ressourceType.id }}" class="btn btn-sm btn-default">
+              <div class="btn btn-sm btn-default">
```
MattiSG: Shouldn't that be a `<label>`?

test/integration/base-suite/2 - Fill demandeur scenario.js
```diff
@@ -0,0 +1,13 @@
...
+steps: [
+    HomepageComponent.start(),
+    {
+        'IndividualFormComponent.birthDateInput': true
+    },
+    IndividualFormComponent.setBirthDateInput(DEMANDEUR_BIRTHDATE),
+    IndividualFormComponent.submit(),
+    {
+        'RecapComponent.firstPersonBirthdate': DEMANDEUR_BIRTHDATE_PLAIN_TEXT
+    }
```
MattiSG: > Try to keep dangling commas, as done in later scenarios.

test/integration/config-ie.json
```diff
@@ -0,0 +1,15 @@
+{
+    "baseURL": "http://localhost:9000",
+    "quit": "on success",
+    "bail": true,
+    "timeout": 10000,
+    "seleniumServerURL": {
+        "hostname" : "ondemand.saucelabs.com",
+        "port" : 80
```
MattiSG: > Maybe add the `auth` part without the key, but with the username, to make it easier to understand how to use it…?

app/styles/front.scss
fpagnoux: This is a ` .resultats-accordion-needs-n-2 i` selector, so maybe not as big you think, but I agree still too big, I just realised there are some other `i` inside the block.

The point of this block is to vertically center the warning sign:

Without it : 
image

With it : 
image

app/views/partials/foyer/capture-montant-ressource.html
fpagnoux: `index` is here to identify the individu, not the year. I would suggest: 
`{{ ressource.type.id }}-month_{{ $index + 1 }}-individu_{{ index }}` and `{{ ressource.type.id }}-year-individu_{{ index }}`.

(The need to have an id per individu may disappear with the resource declaration split).

app/views/partials/foyer/ressources/types.html
fpagnoux: `label` is the enclosing tag.

test/integration/base-suite/2 - Fill demandeur scenario.js
fpagnoux: Expect in config.json 😛 
```sh
undefined:6
}
^
SyntaxError: Unexpected token }
```

test/integration/config-ie.json
fpagnoux: 👍

MattiSG:
GTM

fpagnoux:
Rebasing and merging.

PR #321, opened 2016-07-21
Rendre accessible le choix des types de ressources (1/2)

fpagnoux:
This is based #311 and #303, without the angular upgrade.

Even though it doesn't improve the accordions, it make the resource types buttons more accessible.

Code has been reviewed before in the previous PR, but a double check would be nice to avoid errors brought by the rebase.


MattiSG:
Logos on the homepage have a white background.
capture d ecran 2016-07-21 a 18 32 44

MattiSG:
Fixed after a `grunt build`. Sorry for the false alert.

MattiSG:
GTM

PR #324, opened 2016-07-26
Split resources declaration

### LINKED ISSUE #301: Éclater la déclaration des ressources
En tant qu'utilisateur ayant plusieurs personnes dans mon foyer,
Je peux déclarer les ressources de chaque membre de ma famille indépendamment,
Pour simplifier la saisie

fpagnoux:
Fixes #301 

The main tricky thing here is the navigation between the resources declaration of the family members. 

It is based on substates and scope nesting. The parent state, `/ressources/`, 
 has a controller with a few functions and properties that are available in children states. 

For instance, in `/ressources/0/types/`'s controller, we call`$scope.declareNextIndividuResources`, a function defined in the  `/ressources/` controller.

I don't know how angular-idiomatic it is, but this is already used with the parent state `foyer`, and I preferred it to firing events. 

Also, the `yes-no-question` directive was improved to handle initial value, and to work with `ng-repeat`.


fpagnoux:
Ready for review :). The changelog is big, but a love of stuff have been moved.

MattiSG:
Move the “add more resources” bottom button to the left, to decrease risk of misclicks, make the “next” action more obvious, and use spatial cues for “back”.

capture d ecran 2016-07-28 a 13 40 14

It should probably also be de-emphasized (grey rather than green), as the user has just been through selecting the resources. Going back is an error case, not the standard.

app/js/controllers/foyer/recapSituation.js
```diff
@@ -181,25 +150,16 @@ angular.module('ddsCommon').controller('RecapSituationCtrl', function($scope, $s
...
         buildYm2Recap();
     }
+
+    if ($scope.situation.patrimoine && $scope.situation.patrimoine.captured) {
+        buildRecapPatrimoine();
+    }
+
+    $scope.$on('patrimoineCaptured', function() {
+        buildRecapPatrimoine();
+    });
```
MattiSG: > This is equivalent to `$scope.$on('patrimoineCaptured', buildRecapPatrimoine)` (unless `buildRecapPatrimoine` is dynamically assigned, which it doesn't seem to be.

app/js/controllers/foyer/ressources/enfants.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+angular.module('ddsApp').controller('FoyerRessourcesEnfantsCtrl', function($scope, SituationService, RessourceService) {
+    var momentDebutAnnee = moment($scope.situation.dateDeValeur).subtract('years', 1);
+    $scope.debutAnneeGlissante = momentDebutAnnee.format('MMMM YYYY');
+
+    $scope.enfants = SituationService.getEnfants($scope.situation);
+    $scope.enfants.forEach(function(enfant) {
+    	enfant.hasRessources = RessourceService.getMainScreenRessources(enfant).length > 0 || false;
```
MattiSG: Setting the default value of a Boolean to `false` seems pretty weird, as the only way the `false` could be assigned would be if the leftmost part of the expression was itself `false`…

app/js/directives/yesNoQuestion.js
```diff
@@ -8,35 +8,44 @@ angular.module('ddsApp').directive('yesNoQuestion', function() {
...
+            }
+
+            var propertyPath = $attributes.model,
+                parsedPropertyPath = parse(propertyPath),
+                targetProperty = parsedPropertyPath.targetProperty,
+                targetObject = $scope.$eval(parsedPropertyPath.targetObject);
 
-    var parsedPropertyPath, targetObject;
+            // We need a unique name for the radio input, in case this directive is used in a ng-repeat.
+            $scope.uniqueFieldName = 'field.' + Math.random().toString(36).slice(2);
```
MattiSG: Why `slice`?

MattiSG:
I'm really happy with the end result, I think it makes much more sense this way.

However, there are too many code changes for a proper review, especially considering how so many are just moving code around.
I had a quick read through and wrote a few remarks on stuff that sprung up, but I can't pretend to ensure proper algorithmic equivalence or rightness.

Considering this changes the navigation path and how close the different PRs are to merging, I would be in favour of merging #320 first, merging upstream or rebasing this PR over `master`, updating the integration tests and using them as a safety net for checking such large code changes.

app/js/controllers/foyer/ressources/enfants.js
fpagnoux: The default value useless by a refactoring, good catch 👍

app/js/directives/yesNoQuestion.js
fpagnoux: To avoid the `0.` : 
`field.0.6i0ig783tek1q0k9` -> `field.6i0ig783tek1q0k9`

fpagnoux:
@MattiSG :
image

?

MattiSG:
👌

fpagnoux:
Ok, rebasing and adding integration tests.

fpagnoux:
I rebased, and adapted the integration test. 

I also added a link  in the recap to come back to the result page after modifying the situation.

Most users don't modify their situation, but it is very useful for dev, and now there will be potentially many steps to go back to the end of the journey following the normal steps.

image

I'm open to suggestions about wording (Recalculer mes aides ? Recalculer mes résultats ?)

fpagnoux:
@MattiSG : As there was conflictual rebases, a double check of the behavior would be reassuring.

MattiSG:
We shouldn’t add new features to an already long PR. New features ask for new choices, and bring in new bugs.

fpagnoux:
Ok, I rolled it back. Doesn't change the need for post-release double-check though.

MattiSG:
Inconsistent behaviour:

If I have children,
When I say they have resources and then fill in none,
When I edit the resources for adults,
Then I am directed straight to the list of resources of my children,
While I expected to be asked again if they have resources.

This doesn't happen if I edit after having received results.

- - - - -

I don't find this behaviour particularly troubling, but it might be a symptom for a lack of update or a variable updated by several actors. I let you decide whether to act on this or not.

MattiSG:
Ok for behaviour, too much code for proper review, as mentioned before.

Cross-browser tested in:

- Sf9.1.2
- Ch51
- FF45
- IE10
- IE11

GTM

fpagnoux:
Good catch, an intermediate state variable in scope was not updated in that case.

fpagnoux:
Rebasing and merging.

PR #325, opened 2016-07-28
Give focus to a title element on state change

### LINKED ISSUE #289: Donner le focus aux nouveaux champs à la validation
En tant que déficient visuel,
J'ai un retour auditif sur le changement d'écran,
Pour découvrir les nouvelles questions

Flightan:
In order to give feedback to screen reader users we give focus to the main title (the first `<h1>` in the DOM) on every new application state. Fixes #289


app/js/app.js
```diff
@@ -211,6 +211,16 @@ ddsApp.run(function($rootScope, $state, $stateParams, $window, $modalStack, $anc
         }
     });
 
+    $rootScope.$on('$stateChangeSuccess', function () {
+        // Default - set page focus to h1
+        $timeout(function() {
```
MattiSG: Why make this async?

app/views/partials/a-propos.html
```diff
@@ -1,5 +1,5 @@
 <div class="container">
-  <h1>A propos</h1>
+  <h1 tabindex="-1">A propos</h1>
```
MattiSG: Can't these be added dynamically through JS to avoid repeating (and forgetting to add) the attribute on each page?

app/views/partials/foyer/individu-form.html
```diff
@@ -13,7 +13,6 @@
         date-autocomplete
         before-today
         date-naissance-max-age="options.maxAge"
-        auto-focus
```
MattiSG: This feels like a regression, at least for sighted users.

MattiSG:
If the vocalisation of the autofocused birth date is bad (“date de naissance”), I'd rather improve the 
`<label>` by making it more contextual, like we do in many other screens (“votre date de naissance”, “date de naissance de votre conjoint”…), over asking users to select the field.

app/views/partials/a-propos.html
Flightan: I chose not to do it to handle it differently in some cases (eg. focus on birth date). I'll add it dynamically and add some sort of exception.

app/js/app.js
Flightan: This is to add it to the next angular digest cycle, waiting for the page to load before executing the call.

MattiSG: :bowel:

fpagnoux:
The drawback of focusing on `h1` is that we have unwanted graphic effects: 

image

Flightan:
@fpagnoux because I forgot to uncomment `+// h1[tabindex="-1"] {
 +//   outline: none;
 +// }`

MattiSG:
That should be removable through CSS.

fpagnoux:
Tested and approved :+1:

app/js/app.js
```diff
@@ -211,6 +213,22 @@ ddsApp.run(function($rootScope, $state, $stateParams, $window, $modalStack, $anc
...
+            // $timeout adds the function to the next angular digest cycle,
+            // waiting for the page to load before executing it
+            $timeout(function() {
+                var title = document.querySelector('h1');
+                // If anyone wants to set a tabindex manually, do not overwrite it
+                if (title && title.tabIndex < 0) {
+                    title.tabIndex = -1;
+                    title.focus();
+                }
+            }, 0);
```
MattiSG: > For Angular's `$timeout`, the `delay` parameter is optional and defaults to `0`.

app/js/app.js
```diff
@@ -211,6 +213,22 @@ ddsApp.run(function($rootScope, $state, $stateParams, $window, $modalStack, $anc
...
 
+    $rootScope.$on('$stateChangeSuccess', function (event, current) {
+        // Default - set page focus to first h1
+        if (! current.preventFocus) {
+            // $timeout adds the function to the next angular digest cycle,
+            // waiting for the page to load before executing it
+            $timeout(function() {
+                var title = document.querySelector('h1');
+                // If anyone wants to set a tabindex manually, do not overwrite it
+                if (title && title.tabIndex < 0) {
```
MattiSG: `(undefined < 0) === false`, how can this condition focus an element with no `tabindex` set?

MattiSG: Because the [default value is `-1`](https://html.spec.whatwg.org/multipage/interaction.html#dom-tabindex), even though it has to be _explicitly set to `-1`_ to support `.focus()`… Ok, thanks, DOM API designers 😑

MattiSG:
@Flightan is off, I take ownership of this PR 🎏

PR #328, opened 2016-08-01
Make "pensions alimentaires versées" page accessible

fpagnoux:
Connected to #291 
Connected to #166

### DIFF
--- a/app/js/controllers/foyer/pensionsAlimentaires.js
+++ b/app/js/controllers/foyer/pensionsAlimentaires.js
@@ -65,17 +65,19 @@ angular.module('ddsApp').controller('FoyerPensionsAlimentairesCtrl', function($s
         RessourceService.spreadIndividuRessources(individu, months, ressource, $scope.situation.dateDeValeur);
     };
 
-    $scope.submit = function() {
-        $scope.individusVM.forEach(function (individuVM) {
-            // Remove old pensions alimentaires versees
-            individuVM.individu.ressources = _.filter(individuVM.individu.ressources, function(ressource) {
-                return ! _.contains(['pensionsAlimentairesVersees'], ressource.type);
+    $scope.submit = function(form) {
+        form.submitted = true;
+        if (form.$valid) {
+            $scope.individusVM.forEach(function (individuVM) {
+                // Remove old pensions alimentaires versees
+                individuVM.individu.ressources = _.filter(individuVM.individu.ressources, function(ressource) {
+                    return ! _.contains(['pensionsAlimentairesVersees'], ressource.type);
+                });
+                if ($scope.situation.parentsPayPensionsAlimentaires) {
+                    applyPensionToIndividu(individuVM);
+                }
             });
-            if ($scope.situation.parentsPayPensionsAlimentaires) {
-                applyPensionToIndividu(individuVM);
-            }
-        });
-
-        $scope.$emit('pensionsAlimentaires');
+            $scope.$emit('pensionsAlimentaires');
+        }
     };
 });
--- a/app/js/controllers/foyer/pensionsAlimentaires.js
+++ b/app/js/controllers/foyer/pensionsAlimentaires.js
@@ -41,7 +41,7 @@ angular.module('ddsApp').controller('FoyerPensionsAlimentairesCtrl', function($s
                 onGoing: true
             }
         };
-    };
+    }
 
     var demandeur = _.find($scope.situation.individus, { role: 'demandeur' });
     var conjoint = _.find($scope.situation.individus, { role: 'conjoint' });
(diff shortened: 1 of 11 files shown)


MattiSG:
“Connected to” or “fixes”?

app/js/controllers/foyer/pensionsAlimentaires.js
```diff
@@ -24,13 +24,13 @@ angular.module('ddsApp').controller('FoyerPensionsAlimentairesCtrl', function($s
             })
             .value();
 
-        return result || 0;
+        return Math.round(result) || 0;
```
MattiSG: Can't this be frustrating for the user if the value entered had decimals and they are systematically removed?

(sorry, I can't figure out where the data is coming from)

app/js/controllers/foyer/pensionsAlimentaires.js
```diff
@@ -51,31 +49,29 @@ angular.module('ddsApp').controller('FoyerPensionsAlimentairesCtrl', function($s
         $scope.individusVM.push(createIndividuVM(conjoint));
     }
 
-    $scope.hasPensionsAlimentaires = false;
+    $scope.situation.hasPensionsAlimentaires = false;
```
MattiSG: What about renaming that variable? `has` is always ambiguous between “receives” and “gives” (there's probably a better verb than “give”).

MattiSG:
I fail to see how the `round`ing fixes #166.

app/js/controllers/foyer/pensionsAlimentaires.js
fpagnoux: It could be, and this was a broad behaviour of mes-aides. Changing it to rounding to cents.

app/js/controllers/foyer/pensionsAlimentaires.js
fpagnoux: Usually you pay for child support. Changing it to `situation.parentsPayPensionsAlimentaires`

app/js/controllers/foyer/pensionsAlimentaires.js
MattiSG: It definitely was: #94.

fpagnoux:
#166 was mentioning two problems:
- Too many decimals in year total. This is solved by rounding.
- Year total incremented when resubmitting the situation. Fixed [here]( https://github.com/sgmap/mes-aides-ui/blob/e028d5fda3da05da8a350f863c645ecaf1fcdcd2/app/js/controllers/foyer/pensionsAlimentaires.js#L69) in e028d5fda3da05da8a350f863c645ecaf1fcdcd2.

Bug: We would remove the previously declared pensions alimentaires versées only if the user declared he paid none. So if the user was maintaining he was paying one, we would add the ones recorded in the situation to the one "re-declared".

fpagnoux:
I thought we were avoiding `fixes` as it automatically close issues when merging the PR.

MattiSG:
> automatically close issues when merging the PR

Right, I had forgotten about the “communication” phase that has to be manual due waffleio/waffle.io#616. Sorry for the noise.

app/js/controllers/foyer/pensionsAlimentaires.js
```diff
@@ -51,34 +50,34 @@ angular.module('ddsApp').controller('FoyerPensionsAlimentairesCtrl', function($s
...
 
-    $scope.submit = function() {
-        $scope.individusVM.forEach(function(individuVM) {
-            if ($scope.hasPensionsAlimentaires) {
-                ['pensionsVersees'].forEach(function(typePension) {
-                    applyPensionToIndividu(individuVM, typePension);
-                });
-            } else {
+    $scope.submit = function(form) {
+        form.submitted = true;
```
MattiSG: This is added but there's no addition of this variable in the view. Is it normal?

fpagnoux: [This](https://github.com/sgmap/mes-aides-ui/blob/master/app/views/partials/foyer/capture-montant-ressource.html#L51) is included in the view ;)

MattiSG:
Unfortunately rounding is still not very proper for user display, since trailing `0` are removed…

capture d ecran 2016-08-02 a 18 50 04

The proper implementation should be with `toFixed`, but that breaks because the resulting value is a `String` rather than a `Number`. We probably shouldn't handle this in this PR.

MattiSG:
I don't understand why 2f7f8f4d2d8649eed8288842313986a28130ca6d isn't needed… it doesn't seem to change anything. I'll let you decide whether it should be included or not.

MattiSG:
GTM

PR #329, opened 2016-08-01
Upgrade angular to make accordions accessibles

fpagnoux:
Connected to #300


app/views/partials/droits-eligibles-list.html
```diff
@@ -1,9 +1,8 @@
-<accordion close-others="false">
-  <accordion-group is-open="droit.open"
-               class="resultats-accordion"
-               ng-repeat="(droitId, droit) in list|orderObjectBy:'label'"
-               ng-class="{'resultats-accordion-needs-n-2': shouldDisplayYM2Warning(droit), 'panel-opened': droit.open}">
-    <accordion-heading>
+<uib-accordion close-others="false">
+  <uib-accordion-group is-open="droit.open"
+     class="resultats-accordion {{ getAccordionClass(droit) }}"
```
MattiSG: Could this be made more readable by using an object `ng-class` syntax, like:

```
{ resultats-accordion: true,
  resultats-accordion-needs-n-2: shouldDisplayYM2Warning(),
  panel-opened: droit.open }
```

?

MattiSG: > Using a more generic class name (like “results”) and using a selector like `uib-accordion-group.results` carries as much information while avoiding repetition of tag implementation in class names.

test/integration/base-suite/5 - Declare being SDS scenario.js
```diff
@@ -10,7 +10,7 @@ steps: [
     },
     LogementFormComponent.setZipCodeInput('61500'),
     {
-        'LogementFormComponent.cityInput': '0',
+        'LogementFormComponent.city': /AUNAY/,
```
MattiSG: > Increase forward compatibility by making the RegExp case-insensitive (`/AUNAY/i`).

bower.json
```diff
@@ -21,23 +21,23 @@
...
-    "angular": "~1.2.29",
-    "angular-sanitize": "~1.2.29",
-    "angular-animate": "~1.2.29",
+    "angular": "~1.4",
+    "angular-sanitize": "~1.4",
+    "angular-animate": "~1.4",
     "angular-ui-router": "~0.2.10",
     "angular-uuids": "~0.0.3",
-    "angular-bootstrap": "0.12.0",
+    "angular-bootstrap": "1.3",
```
MattiSG: Latest version is `2.0.1`. If paying upgrade costs, why not use v2?

MattiSG:
According to [Angular Bootstrap’s migration guide](https://github.com/angular-ui/bootstrap/wiki/Migration-guide-for-prefixes), here are a few other renames that don't seem to have been taken into account. These classes should probably be updated,  **if we're sure they used to target the Angular-Bootstrap elements**:

- [ ] `.alert` → `.uib-alert`.
- [ ] In `app/styles/_common.scss#27`: `.carousel`.
- [ ] In `app/styles/_common.scss#27`: `.pagination` (but it doesn't seem to be used anywhere).
- [ ] In `app/styles/_theme.scss`: `.dropdown` (but it doesn't seem to be used anywhere).
- [ ] In `app/styles/_theme.scss`: `.pagination` (but it doesn't seem to be used anywhere).
- [ ] In `app/styles/_theme.scss`: `.pager` (but it doesn't seem to be used anywhere).
- [ ] In `app/styles/_theme.scss#373`: `.progress` (but it doesn't seem to be used anywhere).
- [ ] In `app/views/partials/foyer/patrimoine.html`: `tooltip` (used as an attribute, unsure of the actual implementation called).

(I searched for all the renames throughout the whole app, so the list above should be exhaustive)

MattiSG:
Can it really be possible that upgrading from Angular 1.2 to 1.4 doesn't have any impact whatsoever on the code?

Here are a few highlights from the [migration guide](https://docs.angularjs.org/guide/migration#migrating-from-1-2-to-1-3) which should be assessed:

- The isolated scope of a component directive no longer leaks into the template that contains the instance of the directive. This means that you can no longer access the isolated scope from attributes on the element where the isolated directive is defined.
- `$setViewValue(value)` - This method still changes the `$viewValue` but does not immediately commit this change through to the `$modelValue` as it did previously. 
- The order of items when using ngRepeat to iterate over object properties was guaranteed to be consistent by sorting the keys into alphabetic order. Now, the order of the items is browser dependent based on the order returned from iterating over the object.
- The `select` directive will now use strict comparison. Non-string scope values (such as Number or Boolean) will not be matched against equivalent option strings.

cc @Flightan

bower.json
fpagnoux: Accordion doesn't work as is with `2.0.1`. It seems to be the only problem.

bower.json
```diff
@@ -21,23 +21,23 @@
...
     "font-awesome": "~4.2"
   },
   "devDependencies": {
-    "angular-mocks": "~1.2.29"
+    "angular-mocks": "~1.4"
   },
   "testPath": "test/spec",
   "resolutions": {
-    "angular": "1.2.29"
+    "angular": ">=1 <1.4"
```
MattiSG: I thought we upgraded to Angular 1.4, not 1.3; also, that `angular-bootstrap` v1.3 needed Angular >=1.4.

MattiSG: Here's what `bower install` yields:

```
Please note that,
    angular-bootstrap#0.12.0 depends on angular#>=1 <1.3.0 which resolved to angular#1.2.30
    angular-sanitize#1.2.30 depends on angular#1.2.30 which resolved to angular#1.2.30
    angular-messages#1.4.2 depends on angular#1.4.2 which resolved to angular#1.4.2
    angular-animate#1.4.12, angular-mocks#1.4.12, angular-sanitize#1.4.12 depends on angular#1.4.12 which resolved to angular#1.4.12
    sgmap-mes-aides-ui depends on angular#~1.4 which resolved to angular#1.4.12
    angular-ui-router#0.2.13 depends on angular#>= 1.0.8 which resolved to angular#1.5.8
    ngstorage#0.3.0 depends on angular#>=1.0.8 which resolved to angular#1.5.8
    angular-bootstrap#1.3.3 depends on angular#>=1.4.0 which resolved to angular#1.5.8
Resort to using angular#>=1 <1.4 which resolved to angular#1.2.30
Code incompatibilities may occur.
```

MattiSG:
Now Angular is _actually_ updated… and we do get new errors.

- [ ] Accordions take ages to open under Chrome.
- [ ] Lodging buttons are not aligned under Safari (see screenshot).
capture d ecran 2016-08-04 a 09 27 13

fpagnoux:
The buttons not being aligned is a bug that also exists on other browsers, before updating angular (I found out about this during integration testing). However, I don't reproduce with  Safari 9.1 -_-

MattiSG:
The issues were due to outdated bower packages that had not been cleaned. I now ensured such packages are cleaned on each installation.

MattiSG:
Passed integration tests on IE10, FF45.
- [ ] Still fails under Ch51 due to long accordions, can't reproduce manually.
- [ ] Date input fails under IE11 ([screencast](https://saucelabs.com/jobs/ea25e7c412dc4a149fc6565780ef4638)).
capture d ecran 2016-08-04 a 12 33 53

MattiSG:
Now passes IE11 and Edge.

MattiSG:
Ok, I can't seem to get the integration tests to pass under Chrome. It seems to interpret the default `.35s` in Bootstrap transition duration as `35s`  😑 
I can't reproduce manually.
I'm using `selenium-server-standalone v2.53.1`, `ChromeDriver 2.22.397929` and `Chrome v51.0.2704.103` under `OS X 10.11.6`.
@fpagnoux, if you can't reproduce either, I'm in favour of considering this mergeable.

I would need a review of my changeset. If we want to rebase this branch, I would prefer that @fpagnoux rebases his own fixups. I'm fine with simply doing an upstream merge.

Gruntfile.js
```diff
@@ -95,70 +94,10 @@ module.exports = function (grunt) {
...
-            describe: false,
-            expect: false,
-            inject: false,
-            it: false,
-            jasmine: false,
-            spyOn: false
-          }
-        }
-      }
-    },
```
fpagnoux: 🎉

app/js/lib/cleave.min.js
```diff
@@ -0,0 +1 @@
+../../../node_modules/cleave.js/dist/cleave.min.js
```
fpagnoux: > We could also add a route to serve this file.

fpagnoux:
Seems GTM to me.

app/js/lib/cleave.min.js
MattiSG: We definitely should. The symlink is weak and adds work for linting, as tools believe a minified element is in the source. I'll change this.

That would remove the need to switch from the jshint grunt plugin to the native version, but as discussed IRL, we prefer to keep the switch in order to get the latest versions and be more standard.

MattiSG:
I still get this slow animations thing… Updated Chrome to `52.0.2743.116`, chromedriver to `2.23`, tried to start through `selenium-server-standalone` or through `chromedriver` directly… nothing changes 😖 yet I can't reproduce on [SauceLabs](https://saucelabs.com/beta/tests/959707308b79474f8a89e74ee309d474).

I am able to avoid this behaviour by starting the driver in another session, thus making it headless. It could thus be linked to GPU acceleration on my MacBook Air, which would explain why it's not reproducible on the MacBook Pro we tested on. I did compare the `chrome://flags` through, on a manually-started browser and on a chromedriver-started one, and both seem equal.

PR #330, opened 2016-08-02
Improve integration tests reliability

### LINKED ISSUE #197: Améliorer la résilience de la saisie de code postal

MattiSG:
- Fix #197.
- Avoid wasting 90s in SauceLabs tests.
- ~~Ensure `npm start` defaults to `production` environment to decrease risks of discrepancies (cf. #322).~~ (removed from this PR, was too big)

### DIFF
--- a/app/views/partials/foyer/logement.html
+++ b/app/views/partials/foyer/logement.html
@@ -117,6 +117,7 @@
               minlength="5"
               maxlength="5"
               pattern="[0-9]{5}"
+              ng-pattern="/[0-9]{5}/"
               title="Cinq chiffres"
               class="form-control"
               id="postal-code"
--- a/app/js/controllers/foyer/logement.js
+++ b/app/js/controllers/foyer/logement.js
@@ -86,6 +86,9 @@ angular.module('ddsApp').controller('FoyerLogementCtrl', function($scope, $http,
     };
 
     $scope.updateCities = function updateCities() {
+        if (! $scope.logement.postalCode)
+            return;  // the user has made the value invalid since we were called
+
         $scope.retrievingCities = true;
 
         $http.get('/api/outils/communes/' + $scope.logement.postalCode)
--- a/test/integration/config-ie.json
+++ b/test/integration/config-ie.json
@@ -1,6 +1,6 @@
 {
     "baseURL": "http://localhost:9000",
-    "quit": "on success",
+    "quit": "always",
     "bail": true,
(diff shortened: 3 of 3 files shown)


index.js
```diff
@@ -23,9 +23,7 @@ module.exports = function(app) {
         app.use(express.static(path.join(__dirname, 'app')));
 
         viewsDirectory += '/app/views';
-    }
-
-    if ('production' === env) {
+    } else {
```
fpagnoux: I'm not clear with the distinction between these two modes.

What happens when I do `grunt serve` ? Will the app be launched in dev mode ?
And with `npm start` you can launch it in prod mode ?

app/views/partials/foyer/logement.html
```diff
@@ -117,6 +117,7 @@
               minlength="5"
               maxlength="5"
               pattern="[0-9]{5}"
+              ng-pattern="/[0-9]{5}/"
```
fpagnoux: 👍

index.js
fpagnoux: Do you think we do need to have these two modes ? From my usage, the only difference I expect between dev and prod is the live-reloading, but maybe specifying  a particular cache behaviour is needed in dev.

fpagnoux:
GTM, if you tell me we do need the distinct `prod` and `dev` modes.

index.js
MattiSG: `grunt serve` defaults to serve in `dev`. Judging by the Grunt file, I guess you could use `grunt serve:dist` to do some equivalent of `grunt build && NODE_ENV=production npm start`. Maybe I'll just replace that with a `prestart` script.
`npm start` used to start the app without any environment, resulting in broken layouts due to the implementation this changeset corrects.

As for needing the two modes, this could benefit from a longer investigation, which I kept away from this PR. The current behaviour is already pretty much limited to adding `live-reload`, but there's some work to be done in order to understand how disabled exactly is `prerender`, as well as understanding routes precedency rules.

I don't think this is the place and time to handle this topic. This changeset already makes it clear everything should default to `prod` environment, making `dev` an opt-in. Let's wait until we need to update the rules.

MattiSG:
My implementation above of defaulting to `production` is broken, since Express itself [adjusts its behaviour](http://expressjs.com/en/advanced/best-practice-performance.html#set-nodeenv-to-production) based on `NODE_ENV`.

PR #334, opened 2016-08-04
Fix vocalisation of child removal button

### LINKED ISSUE #292: Vocaliser correctement le bouton d'effacement d'un enfant
En tant que déficient visuel,
Je peux comprendre l'action associée au bouton de suppression d'un enfant,
Pour ne pas l'actionner par erreur

- - - - -

Actuellement vocalisé "Bouton fois".

MattiSG:
Fix #292.

Tested with VoiceOver.

### DIFF
--- a/app/views/partials/foyer/enfants.html
+++ b/app/views/partials/foyer/enfants.html
@@ -13,7 +13,7 @@
           <div class="panel-body">
             {{ enfant.firstName }}
             <button type="button" class="close" ng-click="removeEnfant(enfant)">
-              <span class="screenReadersOnly">Effacer</span>
+              <span class="screenReadersOnly">Effacer {{ enfant.firstName }}</span>
               <span role="presentation" aria-hidden="true">&times;</span>
             </button>
             <div class="row">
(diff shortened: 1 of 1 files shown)


app/views/partials/foyer/enfants.html
```diff
@@ -13,7 +13,8 @@
           <div class="panel-body">
             {{ enfant.firstName }}
             <button type="button" class="close" ng-click="removeEnfant(enfant)">
-              <span aria-hidden="true">&times;</span>
+              <span class="screenReadersOnly">Effacer</span>
```
fpagnoux: Nice class, right 😎  ?

app/views/partials/foyer/enfants.html
```diff
@@ -13,7 +13,8 @@
...
+              <span aria-role="presentation" aria-hidden="true">&times;</span>
```
fpagnoux: > So is-it a good practise to add `aria-role="presentation"` when using `aria-hidden="true"`?

MattiSG: To be honest, I don't know exactly. I used to think `role="presentation"` was enough, but [it seems not](http://john.foliot.ca/aria-hidden/).

MattiSG: This should be `role`, not `aria-role` :-/

app/views/partials/foyer/enfants.html
MattiSG: We could improve vocalisation by adding the name of the child.

fpagnoux:
Tested after vocalisation improve, GTM.

PR #339, opened 2016-08-08
Fix ludwig view

MattiSG:
Situation recap was not loaded anymore.


MattiSG:
This changeset reverts rounding in the situation recap.

capture d ecran 2016-08-08 a 18 08 11

WIP to fix it is in the `euros-directive` branch. Broken Ludwig is more important than rounding issues in a very rarely seen part.

PR #341, opened 2016-08-08
Ensure OpenFisca packages are always updated

MattiSG:
If the requirements.txt changes the referenced version, it was not updated


MattiSG:
GTM

PR #346, opened 2016-08-11
Utilise le composant `yes-no-question` pour toutes les questions booléennes

### LINKED ISSUE #291: Remplacer les boutons oui / non par des boutons radio
En tant que déficient visuel,
Je peux déterminer l'état de coche des boutons oui / non,
Pour les valider correctement

MattiSG:
Fix #291.

### DIFF
--- a/app/views/partials/foyer/individu-form.html
+++ b/app/views/partials/foyer/individu-form.html
@@ -169,9 +169,9 @@
       >
       <question>
         {{ isDemandeur() ? 'Avez-vous' : 'A-t-il (elle)' }}
-        travaillé au moins 5 ans entre {{ individu.lastJobEndDate.clone().subtract(10, 'years').format('MMMM YYYY') }} et {{ individu.lastJobEndDate.format('MMMM YYYY') }} ?
+        travaillé au moins 5&nbsp;ans entre&nbsp;{{ individu.lastJobEndDate.clone().subtract(10, 'years').format('MMMM YYYY') }} et&nbsp;{{ individu.lastJobEndDate.format('MMMM YYYY') }}&nbsp;?
       </question>
-      <help-block>1825 jours (5 fois 365) couverts par un contrat de travail, en activité ou en congés.</help-block>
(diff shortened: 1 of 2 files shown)


app/views/partials/foyer/individu-form.html
```diff
@@ -162,23 +162,16 @@
       <span class="help-block">Si vous n’avez jamais eu de contrat de travail, laissez ce champ vide.</span>
     </div>
 
-    <div class="form-group" ng-if="individu.lastJobEndDate.year()">
```
MattiSG: The `class="form-group"` has been completely erased. Is it voluntary?

app/views/partials/foyer/patrimoine.html
```diff
@@ -140,11 +136,11 @@ <h2 class="panel-title">
         </div>
 
         <div class="form-group">
-          <label class="col-sm-4 control-label">Détenez-vous un <abbr title="Plan Épargne Logement">PEL</abbr>, un <abbr title="Plan d'épargne retraite populaire">PERP</abbr>, des titres, une assurance-vie, des parts d'une <abbr title="Société d'Investissement à CApital Variable">SICAV</abbr> ou d'un <abbr title="Fonds Commun de Placement">FCP</abbr> ?</label>
```
MattiSG: Here, it was only `label`/`button` couple replaced, not the `form-group` altogether. This is inconsistent with the other parts.

app/views/partials/foyer/individu-form.html
MattiSG: Yes: the `<yes-no-question>` directive root is a `<fieldset class="form-group">`, so there's no need to repeat it on the outside of the component.

app/views/partials/foyer/patrimoine.html
MattiSG: Good catch! This indeed broke the layout.

As in reviewed code
capture d ecran 2016-08-12 a 10 51 49

After removing the unnecessary `form-group`:
capture d ecran 2016-08-12 a 10 52 48

PR #350, opened 2016-08-12
Corrige la vocalisation des types de ressources

MattiSG:
Avant :
capture d ecran 2016-08-12 a 11 49 56

Après :
capture d ecran 2016-08-12 a 11 49 19

### DIFF
--- a/app/styles/front.scss
+++ b/app/styles/front.scss
@@ -616,22 +616,17 @@ input:checked + .btn, input:checked + .btn:hover, input:checked + .btn:active:ho
 .button-grid {
   display: flex;
   flex-wrap: wrap;
+  margin-left: -.25em;
+  margin-right: -.25em;
 
   label {
-    width: 33.33%;
-    padding: 0.3em;
+    width: calc(33.33% - .5em);
+    margin: .25em;
+    padding: .25em;
     font-size: 100%;
-
-    .btn {
-      display: flex;
-      align-items: center;
-      justify-content: center;
-      width: 100%;
-      height: 100%;
-      border-top: none;
-      border-bottom: none;
-      font-size: 100%;
-    }
+    display: flex;
+    align-items: center;
+    justify-content: center;
   }
 }
 
--- a/app/views/partials/foyer/ressources/types.html
+++ b/app/views/partials/foyer/ressources/types.html
@@ -20,16 +20,14 @@
               </i>
             </uib-accordion-heading>
             <div class="button-grid">
-              <label
-                ng-repeat="ressourceType in ressourceTypesByCategories[category.id]| orderBy:['positionInList','label']">
-                <input
-                  type="checkbox"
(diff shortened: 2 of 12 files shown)


app/views/partials/foyer/ressources/types.html
```diff
@@ -14,22 +14,23 @@
             is-open="status.open"
             ng-init="status.open = shouldInitiallyOpen(category)">
             <uib-accordion-heading>
-              {{ category.label }}
-              <i class="pull-left fa" aria-hidden="true"
+              <i class="pull-left fa"
```
MattiSG: > Is `.pull-left` still needed if you move the `<i>` to the left?

app/views/partials/foyer/ressources/types.html
```diff
@@ -14,22 +14,23 @@
...
+                ng-class="{ active: selectedRessourceTypes[ressourceType.id] }">
+                <input type="checkbox"
                   ng-model="selectedRessourceTypes[ressourceType.id]"
-                />
-                <div class="btn btn-sm btn-default">
+                  class="screenReadersOnly"
+                  >
                   {{ ressourceType.label }}
-                </div>
+                </input>
```
MattiSG: `<input>` is an empty element, there's no point in closing it here.

MattiSG:
f190493 goes beyond the scope of this PR, but ok for keeping it here since it does improve vocalisation.

app/views/partials/foyer/ressources/types.html
MattiSG: No, and it does actually improve rendering to remove it!

Current:
capture d ecran 2016-08-12 a 14 40 47

Without `pull-left`:
capture d ecran 2016-08-12 a 14 40 55

fpagnoux:
I find it clearer with a little space between the boxes: 
image

@Flightan, what do you think ? 

(The former styling was dead code with this PR, I re-integrated and adapted it)

Flightan:
Good for me

app/styles/front.scss
```diff
@@ -616,22 +616,17 @@ input:checked + .btn, input:checked + .btn:hover, input:checked + .btn:active:ho
 .button-grid {
   display: flex;
   flex-wrap: wrap;
+  margin-left: -2px;
+  margin-right: -2px;
 
   label {
-    width: 33.33%;
+    width: calc(100% * 1/3 - 4px);
```
MattiSG: > Partial support in Android Browser 4.4 refers to the browser lacking the ability to multiply and divide values.
([source](http://caniuse.com/#search=calc))
We can get 1% more support by using `33% - 4px` rather than `100% * 1/3`.
I think this `4px` should rather be something like `.5em` to make the spaces relative to the font size.

MattiSG:
Passes integration tests on Chrome.

PR #351, opened 2016-08-16
Convert homepage definitions to native html elements

Flightan:
The objective of this technique is to provide the definitions of words or phrases by presenting them in a definition list. (https://www.w3.org/TR/WCAG20-TECHS/H40.html)

This removes some ugly clearfix hacks and is overall easier to read IMO. It should be better for accessibility but I don't know how to navigate through html elements using the keyboard.

### DIFF
--- a/app/views/partials/homepage.html
+++ b/app/views/partials/homepage.html
@@ -49,17 +49,10 @@
           <img ng-src="/img/{{ provider.imgSrc }}" alt="{{ provider.label }}"/>
         </div>
         <div class="col-md-10">
-          <div ng-repeat="droit in provider.prestations">
-            <div class="clearfix" ng-if="$even"></div>
-            <div class="col-md-6">
-              <div class="row">
-                <div class="col-xs-10">
-                  <h3>{{ droit.label }}</h3>
-                </div>
-              </div>
-              <p>{{ droit.description }}</p>
-            </div>
-          </div>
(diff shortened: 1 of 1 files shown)


app/views/partials/homepage.html
```diff
@@ -72,22 +65,17 @@ <h2 class="text-center">Simulez ces aides locales</h2>
...
+            <dt>{{ droit.label }}</dt>
+            <dd>{{ droit.description }}</dd>
+          </dl>
+
           <div class="col-md-12 showDetailsButtonBlock">
             <button class="btn" type="button" ng-show="!showDetails && countPrestations(provider) > 2" ng-click="showDetails = ! showDetails">
               Voir les {{ countPrestations(provider) - 2 }} autres aides calculées
             </button>
           </div>
+        </div>
```
Flightan: Bonus missing closing div

fpagnoux:
To navigate non necessarily focusable elements, use:

 >**Voice Over Key + →** 

 **Voice Over Key** is **Caps lock** by default, you can custom it in the `General` tab of the VoiceOver Utility.

[More VO shortcuts](http://axslab.com/articles/ios-voiceover-gestures-and-keyboard-commands.php).

fpagnoux:
Otherwise, it is more semantic and simpler, so GTM for me 👍

MattiSG:
:tada:

PR #355, opened 2016-08-23
Use Bootstrap class for screenReadersOnly CSS

MattiSG:
@fpagnoux was there any specific reason for introducing that class rather than using Bootstrap's default one?

### DIFF
--- a/app/styles/front.scss
+++ b/app/styles/front.scss
@@ -579,16 +579,6 @@ legend {
   font-size: inherit;
 }
 
-.screenReadersOnly {
-  border: 0;
-  clip: rect(0 0 0 0);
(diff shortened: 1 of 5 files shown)


fpagnoux:
Didn't know it existed. Is the implementation the same ? Did you check the behaviour ?

MattiSG:
[Same implementation](https://github.com/twbs/bootstrap/blob/4e163001358170c747e3443982e15dc41d3dcabe/less/scaffolding.less#L125-L134), behaviour is the same with VoiceOver.

MattiSG:
@fpagnoux GTM ?

fpagnoux:
GTM

PR #390, opened 2016-10-17
Redéployer openfisca dans le script de déploiement continu

fpagnoux:
Cette PR permet de
- Ajouter le script de déploiement utilisé en production dans le repo pour qu'il soit versionné.
    - Avant, on n'avait aucune sauvegarde des changements.
- Redéployer openfisca dans le script de déploiement continu, pour garantir qu'il n'y ait pas d'incohérence entre mes-aides et openfisca.
    - Avant, il fallait se connecter avec un user spécifique `openfisca-mes-aides` pour déployer openfisca, avec un risque d'oubli.

Bien sûr, ce qu'il faudrait faire c'est merger complètement les processus de déploiement de la prod et des instances de staging (qui gèrent déjà à la fois openfisca et le front). Mais il reste quelques obstacles, notamment le fait utiliser node 0.10, qui est incompatible avec pm2.

### DIFF
--- a/deploy-prod.sh
+++ b/deploy-prod.sh
@@ -2,6 +2,8 @@
 # This script is not currently meant to be executed from within the repository.
 # Copy or symlink it to some deployment directory.
 
+OPENFISCA_PORT=12000
+
 set -ex
 
 cd /var/www/dds
@@ -15,7 +17,7 @@ grunt build
 
 # Update openfisca
 npm run install-openfisca
-cat openfisca/api_config.ini | sed s/"port = 2000"/"port = 12000"/ > ~/production.ini
+cat openfisca/api_config.ini | sed "s/port = 2000/port = $OPENFISCA_PORT"/ > ~/production.ini
 
 # Restart mes-aides
 sudo restart dds
--- a/deploy-prod.sh
+++ b/deploy-prod.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+# This script is not currently meant to be executed from within the repository.
+# Copy or symlink it to some deployment directory.
+
+set -ex
+
+cd /var/www/dds
+
+# Update mes-aides
+git fetch origin
+git rebase origin/master
+npm install
+grunt build
+
+# Update openfisca
+npm run install-openfisca
+cat openfisca/api_config.ini | sed s/"port = 2000"/"port = 12000"/ > ~/production.ini
+
+# Restart mes-aides
+sudo restart dds
+
+# Restart openfisca
+sudo start openfisca || sudo restart openfisca
+
+echo "Déploiement effectué"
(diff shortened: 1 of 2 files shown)


fpagnoux:
Une fois que c'est mergé:
- [ ] Déprécier l'utilisateur `openfisca-mes-aides`.

deploy-prod.sh
```diff
@@ -0,0 +1,25 @@
...
+# This script is not currently meant to be executed from within the repository.
+# Copy or symlink it to some deployment directory.
+
+set -ex
+
+cd /var/www/dds
+
+# Update mes-aides
+git fetch origin
+git rebase origin/master
```
MattiSG: No, that's not what we want.
We want the deployment to be bullet-proof, and git will complain and stop if, for example, some random file is created (repo is not clean), or if there was a force push at some point (exactly what you can do under stress, which would then break the CD, leading to more stress and more harmful behaviour).

You need to [force a pure remote mirror](https://github.com/sgmap/paie-api/blob/master/deploy.sh#L19-L23).

deploy-prod.sh
```diff
@@ -0,0 +1,25 @@
...
+npm install
+grunt build
+
+# Update openfisca
+npm run install-openfisca
+cat openfisca/api_config.ini | sed s/"port = 2000"/"port = 12000"/ > ~/production.ini
```
MattiSG: The double quotes here are misleading, as they will be interpreted by the shell, but it looks like they are a `sed` match. I recommend you escape the whole expression: `"s/port = 2000/port = 12000/"`.

deploy-prod.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# Restart mes-aides
+sudo restart dds
+
+# Restart openfisca
+sudo start openfisca || sudo restart openfisca
```
MattiSG: Why is it not the same for `dds`?

deploy-prod.sh
```diff
@@ -0,0 +1,25 @@
...
+
+set -ex
+
+cd /var/www/dds
+
+# Update mes-aides
+git fetch origin
+git rebase origin/master
+npm install
+grunt build
```
MattiSG: Why is there still a dependency on `grunt`? Shouldn't this be some NPM script?

deploy-prod.sh
MattiSG: The port and destination file should be `$VARIABLES` defined at the top of the script, to increase the discoverability of the configuration points.

deploy-prod.sh
fpagnoux: No idea. There are some differences between the `init.d` services: dds is multi-instance, while openfisca has only one gunicorn process that then manages several openfisca instances.

I don't know if we should really investigate that now. As I said in my first comment, I think this whole mechanism should be replaced by something we understand totally, similar to what we use for staging. The intended scope of this PR was just to solve some concrete issues that may cause bugs in deployment (like we had yesterday for 10min), and we will definitely need later more work on the production.

MattiSG: What I meant was fixed in https://github.com/sgmap/mes-aides-ui/pull/390/commits/fcfad8e3a322d317c2d08a9c67ec500fdbe337b5 ;)

PR #513, opened 2017-03-19
Fix expected text in integration test

guillett:
* Rewording in f456e6416 was not replicated in tests

Using text from latest version cf. https://github.com/sgmap/mes-aides-ui/commit/4b7fb397b85177b47aaa48dbb38ac4f14ec58cc1#diff-e587a849da685f899e818a7c211889dbR150


test/integration/base-suite/ResultsFixture.js
```diff
@@ -1,5 +1,5 @@
 PRESTATION_TO_TEST_POSITION = 1
 PRESTATION_NAME = /Prime d’activité/
-PRESTATION_DESCRIPTION = /revenus modestes/
+PRESTATION_DESCRIPTION = /complète les revenus/
```
MattiSG: Améliorer la résilience et éviter ce type de modifications dans le futur : ne matcher que `revenus`.

PR #526, opened 2017-03-30
Implement integration test automation with CircleCI and SauceLabs

guillett:


circle.yml
```diff
@@ -0,0 +1,54 @@
...
+    - npm install saucelabs@0.1.1 # allow sending Watai results to SauceLabs
+    - wget https://saucelabs.com/downloads/sc-latest-linux.tar.gz
+    - tar -xzf sc-latest-linux.tar.gz
+    - sudo pip install --upgrade pip
+    - sudo pip install gunicorn
+    - npm run install-openfisca
+
+test:
+  pre:
+    - npm test
```
MattiSG: This shouldn't be in the `pre` section: this is an actual test, not a preparation of the test.

test/integration/config-ci.js
```diff
@@ -0,0 +1,16 @@
...
+    seleniumServerURL: {
+        hostname    : 'ondemand.saucelabs.com',
+        port        : 80,
+    },
+    driverCapabilities: {
+        platform            : 'Windows 7',
+        'tunnel-identifier' : 'circle-' + process.env.CIRCLE_PROJECT_USERNAME + '-' + process.env.CIRCLE_PROJECT_REPONAME + '-' + process.env.CIRCLE_BUILD_NUM + '-' + process.env.CIRCLE_NODE_INDEX
+    },
+    tags        : [ 'circle-ci', '#' + process.env.CIRCLE_BUILD_NUM ],
+    views       : [ 'Verbose', 'SauceLabs' ],
```
MattiSG: This is currently ignored because the original `config.json` also defines views, which in turn prevents the SauceLabs from being active, thus preventing build status to be sent to SauceLabs 😞

Two possible fixes:

1. Remove `views` from the original `config.json`.
2. Enforce `views` through the `--config` CLI switch in `circle.yml`.

guillett: The last line of test.pre (`mkdir $HOME/.watai && cp $WATAI_TESTS/config-ci.js $HOME/.watai/config.js`) should be doing the trick.

guillett: You are right. I have done some amendments to circle.yml.

MattiSG: No, due to [precedence rules](https://github.com/MattiSG/Node-ConfigLoader#paths).

MattiSG: See https://github.com/openfisca/legislation-explorer/pull/71#discussion_r108738852.

guillett: Thanks I used a different strategy `rm $WATAI_TESTS/config.json` and `cp config-ci.js config.js`. I have a passing build in CircleCI and tests are also passing in SauceLabs (and shown as passed there).

MattiSG: Ok. Just have to keep in mind that any change in `config.json` will thus be ignored in CI.

MattiSG:
But IIRC you deactivated multiple containers builds yesterday, so we can't test this anymore 🤔

guillett:
I agree, I deactivated containers because it wasn't useful at the time. Moving back to multiple containers is the easy bit. I just triggered a new build with 3 containers.

To me, the main point of the PR is to have integration test run systematically.
Cross-browser testing is a bonus.

guillett:
The current pipeline does not seem to work with Firefox. https://circleci.com/gh/sgmap/mes-aides-ui/1580#tests/containers/1

At the end of the simulation, clicks on *prestation* headers are not triggering anything. I haven't investigated much.

Can we merge with a single node, running tests on Chrome and delay the cross-browser validation ?

PR #527, opened 2017-03-30
Change "+" to "%20" in cil's mailto

monbocal:
Self explaining ; that was bugging me. Spaces in email's body are now spaces, not plus signs.


MattiSG:
@guillett tu as le droit de merge dans ce cas :wink:

MattiSG:
Je m'en occupe.

PR #537, opened 2017-04-12
Pull out our social network statement

monbocal:
Some rewriting is requiered before going public.


MattiSG:
I don't think there's any strong risk associated with the current wording. If there are improvements to be made, then let's make them, but let's not just remove content :wink:

PR #556, opened 2017-05-12
Rely NVM node default version in deployment script

guillett:


guillett:
Source nvm as per indicated at
https://github.com/creationix/nvm#install-script

guillett:
No longer required as dedicated server was reset.

PR #561, opened 2017-05-18
Improve error reporting on result page

guillett:
* Add error middleware and show formatted JSON in error report

capture d ecran 2017-05-18 a 12 23 54

Helps #559

### DIFF
--- a/app/js/controllers/resultat.js
+++ b/app/js/controllers/resultat.js
@@ -19,14 +19,14 @@ angular.module('ddsApp').controller('ResultatCtrl', function($scope, $rootScope,
     .then(function(situation) {
         $scope.situation = situation;
         return ResultatService.simulate($scope.situation);
-        })
+    })
     .then(function(droits) {
         $scope.droits = droits.droitsEligibles;
         $scope.droitsInjectes = droits.droitsInjectes;
         $scope.noDroits = _.isEmpty($scope.droits.prestationsNationales) && _.isEmpty($scope.droits.partenairesLocaux);
     })
     .catch(function(error) {
-        $scope.error = (error && error.data) || error || true;
+        $scope.error = JSON.stringify((error && error.data), null, 2);
         $scope.encodedError = encodeURIComponent($scope.error);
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -62,4 +62,8 @@ module.exports = function(app) {
     app.route('/*').get(function(req, res) {
         res.sendFile(viewsDirectory + '/front.html');
     });
+
+    app.use(function (err, req, res) {
+      if (err) res.status(err.code || 500).send(err);
```
MattiSG: Don't use brackets if you want, but please insert a newline between the condition and the body block if you don't `return`.

app/js/controllers/resultat.js
```diff
@@ -26,7 +26,8 @@ angular.module('ddsApp').controller('ResultatCtrl', function($scope, $rootScope,
         $scope.noDroits = _.isEmpty($scope.droits.prestationsNationales) && _.isEmpty($scope.droits.partenairesLocaux);
     })
     .catch(function(error) {
-        $scope.error = (error && error.data) || error || true;
+        var errorProxy = (error && error.data) || error || true;
```
MattiSG: Poor naming. You're describing what you are using the object for rather than what it is, but its scope is broader than the line right after.

I would not be shocked by not using a temporary variable at all, even more since:

- `(error && error.data) || error` is equivalent to `error && error.data`.
- Using `JSON.stringify` means the encoded `error` value will yield a truthy, even if the original `error` was a falsy (except if it was `undefined`; it would be very weird for `undefined` to be explicitly thrown, but reading the source promises should make it clear whether this is possible).

index.js
```diff
@@ -62,4 +62,8 @@ module.exports = function(app) {
     app.route('/*').get(function(req, res) {
         res.sendFile(viewsDirectory + '/front.html');
     });
+
+    app.use(function (err, req, res) {
+        if (err) return res.status((err.code || 500)).send(err);
```
MattiSG: Double parens :wink:

guillett: Misunderstanding of https://github.com/sgmap/mes-aides-ui/pull/561#discussion_r118875855.

Thanks !

PR #587, opened 2017-06-13
Fix number display

guillett:
In recap and year-moins-2 views.

Before

capture d ecran 2017-06-13 a 10 14 34

After
capture d ecran 2017-06-13 a 10 14 57

### DIFF
--- a/app/styles/recap-situation.css
+++ b/app/styles/recap-situation.css
@@ -1,0 +1,62 @@
+.recap-situation {
+    border-top: none;
+    background-color: white;
+}
+
+.recap-situation h3 {
+    margin-top: 0;
+}
+
+.recap-situation .section-header {
+    margin-top: 21px;
+}
+
+.accordion-ressources .panel {
+    border: none;
+    font-size: 12px;
+}
+
+.accordion-ressources .panel-heading {
+    padding: 5px 8px;
+}
+
+.accordion-ressources .panel-title {
+    font-size: 14px;
+}
+
+.accordion-ressources .panel-body {
+    padding: 10px;
+}
+
+.accordion-ressources hr {
+    margin-top: 5px;
+    margin-bottom: 5px;
+}
+
+.accordion-ressources .recap-indemnites {
+    border-left: 7px solid #AC6C82;
+}
+
+.accordion-ressources .recap-pensions {
+    border-left: 7px solid #685C79;
+}
+
+.accordion-ressources .recap-allocations {
+    border-left: 7px solid #DA727E;
+}
+
+.accordion-ressources .recap-revenusActivite {
+    border-left: 7px solid #FFBC67;
+}
+
+.accordion-ressources .recap-autre {
+    border-left: 7px solid #39747B;
+}
+
+.accordion-ressources .recap-rpns {
+    border-left: 7px solid #8366d4;
+}
+
+.accordion-ressources .recap-patrimoine {
+    border-left: 7px solid #455C7B;
+}
--- a/app/styles/recap-situation.css
+++ b/app/styles/recap-situation.css
@@ -1,31 +1,26 @@
-.recap-situation {
-    border-top: none;
-    background-color: white;
-}
-
 .recap-situation h3 {
     margin-top: 0;
 }
 
-.recap-situation .section-header {
-    margin-top: 21px;
+.recap-situation h4:not(.panel-title) {
+    margin-top: 1em;
 }
 
 .accordion-ressources .panel {
     border: none;
-    font-size: 12px;
+    font-size: 85%;
 }
 
 .accordion-ressources .panel-heading {
-    padding: 5px 8px;
+    padding: 0.5em 0.8em;
 }
 
 .accordion-ressources .panel-title {
-    font-size: 14px;
+    font-size: 120%;
 }
 
 .accordion-ressources .panel-body {
-    padding: 10px;
+    padding: 0.8em;
 }
 
 .accordion-ressources hr {
--- a/app/styles/recap-situation.css
+++ b/app/styles/recap-situation.css
@@ -7,7 +7,6 @@
 }
 
 .accordion-ressources .panel {
-    border: none;
     font-size: 85%;
 }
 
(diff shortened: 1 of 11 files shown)


app/styles/recap-situation.css
```diff
@@ -0,0 +1,66 @@
...
+    border-top: none;
+    background-color: white;
+}
+
+.recap-situation h3 {
+    margin-top: 0;
+}
+
+.recap-situation .section-header {
+    margin-top: 21px;
```
MattiSG: Use `em` 🙂
This way, not only will it scale better, it will also read properly like “add two blank lines before that title”.
This is valid for all `px` instructions.

app/styles/recap-situation.css
```diff
@@ -0,0 +1,66 @@
+.recap-situation {
+    border-top: none;
+    background-color: white;
```
MattiSG: Useless: already provided by `whiteframe-z1`.

app/styles/recap-situation.css
```diff
@@ -0,0 +1,66 @@
+.recap-situation {
+    border-top: none;
```
MattiSG: The difference is almost impossible to see. If you confirm, please just remove this instruction — this way we can drop the selector altogether  😃

(with `border-top: none`:
capture d ecran 2017-06-13 a 22 31 02
without:
capture d ecran 2017-06-13 a 22 30 59
)

app/styles/recap-situation.css
```diff
@@ -0,0 +1,66 @@
...
+    background-color: white;
+}
+
+.recap-situation h3 {
+    margin-top: 0;
+}
+
+.recap-situation .section-header {
```
MattiSG: Replace the `section-header` class by `h4`, so we can get rid of this class and selector altogether. If this makes the single `h4` that doesn't have this class look bad, then just make it into a `h5`.

app/styles/recap-situation.css
```diff
@@ -0,0 +1,66 @@
...
+    margin-top: 21px;
+}
+
+.accordion-ressources .panel {
+    border: none;
+    font-size: 12px;
```
MattiSG: Always declare `font-size` in `%` of the parent.

app/styles/recap-situation.css
```diff
@@ -0,0 +1,66 @@
...
+
+.accordion-ressources .panel-title {
+    font-size: 14px;
+}
+
+.accordion-ressources .panel-body {
+    padding: 10px;
+}
+
+.accordion-ressources hr {
```
MattiSG: > These `<hr>` could be properly replaced by a `border-bottom` and a `margin-bottom` on every `.accordion-ressources .row`.

app/styles/recap-situation.css
```diff
@@ -0,0 +1,66 @@
...
+    margin-top: 5px;
+    margin-bottom: 5px;
+}
+
+.accordion-ressources .recap-indemnites {
+    border-left: 7px solid #AC6C82;
```
MattiSG: You can factorise the `7px solid` by not using the shorthand property.

First, use one selector that is valid for all the `.recap-` panels. If there is none, introduce one (possibly replacing `class="recap-indemnites"` by `class="ressource-recap indemnites"`).
In CSS, declare for this class: `border-style: solid; border-width: 7px;` (and put it in `em` along the way).
Then, you only have to specify `border-color` for each colour variant 😃

app/styles/recap-situation.css
```diff
@@ -0,0 +1,66 @@
...
+
+.accordion-ressources .recap-rpns {
+    border-left: 7px solid #8366d4;
+}
+
+.accordion-ressources .recap-patrimoine {
+    border-left: 7px solid #455C7B;
+}
+
+.right-align {
```
MattiSG: Never create such non-semantic helper classes, especially when they are so terribly widely scoped.
It is not that we want some random `div` to be right-aligned. It is that, when we display a number in a list, we want it to be right-aligned. This shouldn't leak in the HTML.
The proper way to do this would be to change the markup altogether, but this is out of the scope of this PR.
At least, let's rename that class: which elements does it represent? I guess it is a `ressource-value`, WDYT?

test/integration/family-suite/90 - Declare YM2 resources scenario.js
```diff
@@ -3,7 +3,7 @@ description: 'Declare YM2 resources',
 steps: [
     ResultComponent.declareYM2Resources(),
     {
-        'YM2ResourcesComponent.suggestionText': /12000 €/,
+        'YM2ResourcesComponent.suggestionText': /12,000.00 €/,
```
MattiSG: This is not the proper notation for currencies in France. The thousands separator should be ` `, and the comma should be `,`.
If you want to use the native `number` filter (which is probably a good idea — unless it should be [`currency`](https://docs.angularjs.org/api/ng/filter/currency) so we don't have to repeat `€`? 😉), then you will need to include [Angular's `i18n`](https://docs.angularjs.org/guide/i18n).

bower.json
```diff
@@ -22,14 +22,15 @@
   ],
   "dependencies": {
     "angular": "~1.4",
-    "angular-sanitize": "~1.4",
     "angular-animate": "~1.4",
-    "angular-ui-router": "~0.2.10",
     "angular-bootstrap": "1.3",
+    "angular-i18n" : "1.6.4",
```
MattiSG: > We'd like to get rid of the `bower` dependency and install everything through `npm`. When you add new frontend dependencies, please try to declare them in `package.json` rather than `bower.json`.
> In this specific case, as an Angular dependency, it might indeed be more pertinent here so that dependency constraints can be solved properly.

MattiSG:
> Take CSS remarks from review into account

Not a [great commit message](https://chris.beams.io/posts/git-commit/#seven-rules) :wink:

> to squash! Make insertion revertable

The syntax is `squash!` if you want to benefit from [`--autosquash`](https://robots.thoughtbot.com/autosquashing-git-commits).

PR #590, opened 2017-06-15
Prevent situation update and chain situations together

### LINKED ISSUE #576: Ne pas permettre la modification d'une situation
Une situation enregistrée ne doit pas être modifiable.

En cas de modification dans l'application, il faudrait indiquer l'ID de la situation originale dans l'attribut `previous_id` de la nouvelle situation par exemple.

guillett:
Fixes #576

Helps investigations for issues similar to #589 (cf. https://github.com/sgmap/mes-aides-ui/issues/576#issuecomment-308729307)

### DIFF
--- a/app/js/services/situationService.js
+++ b/app/js/services/situationService.js
@@ -99,13 +99,13 @@ angular.module('ddsCommon').factory('SituationService', function($http, $session
 
         save: function(situation) {
             return $http({
-                    method: situation._id ? 'put' : 'post',
-                    url: '/api/situations/' + (situation._id || ''),
-                    data: this.createApiCompatibleSituation(situation)
-                }).then(function(result) {
-                    situation._id = result.data._id;
-                    return result.data;
-                });
+                method: 'post',
+                url: '/api/situations/',
+                data: this.createApiCompatibleSituation(situation)
+            }).then(function(result) {
+                situation._id = result.data._id;
+                return result.data;
+            });
         },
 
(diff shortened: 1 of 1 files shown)


app/js/services/situationService.js
```diff
@@ -120,6 +120,7 @@ angular.module('ddsCommon').factory('SituationService', function($http, $session
             }
 
             var result = {
+                referenceId: situation._id,
```
MattiSG: I don't find this name very clear. Pretty much anything could be a “reference”. What about `parent`, `parentSituation`, `previousVersion`, `modifiedFrom`… ?

guillett: I used that specific word to reflect OpenFisca naming convention. I agree with your point but given OpenFisca's definition of reference, it makes sense to be consistent.

guillett:
The commit on the API was ready, the PR was not created. I prefer to keep `PUT` support for a few days to prevent errors for current users.

app/js/services/situationService.js
MattiSG: I'm not sure OpenFisca's “reference” means what you think it means :wink:
See https://github.com/openfisca/openfisca-core/pull/525

MattiSG:
> I prefer to keep PUT support for a few days to prevent errors for current users.

This is scary. Does this mean we don't master our cache invalidation? Could there be any users impacted other than the ones who started a simulation before deployment and finish it after deployment?

PR #592, opened 2017-06-20
Add Y2M form validation and integration tests

### LINKED ISSUE #93: Rendre visible les erreurs à l'entrée des revenus n-2
Sur la page Vos revenus imposables en 2013, si l'on entre un nombre négatif (on peut supposer qu'une personne appuis par erreur sur le - en même temps qu'elle appuis sur le 9), alors deux cas : 
- Quand on appuie sur valider rien ne se passe (à cause d'une erreur 500).
- Quand on appuie sur valider on arrive sur la page "Votre revenu fiscal de référence" (c'est le cas si on a mis une personne à charge) et là quand on appuis sur valider rien ne se passe, et ce même si on a mis une valeur juste sur la page "Votre revenu fiscal de référence" (à cause d'une erreur 500) (et là la personne ne se rendra pas compte du - en trop vu que ce n'est plus la même page).

guillett:
Superseeds #410.
Fixes #93.


app/js/controllers/foyer/yearMoins2.js
```diff
@@ -36,27 +36,29 @@ angular.module('ddsApp').controller('FoyerRessourceYearMoins2Ctrl', function($sc
...
+                    return ressource.periode !== $scope.yearMoins2 && (! _.find(categoriesRnc, { id: ressource.type }));
+                });
 
-            // Remove the empty values from rnc
-            individuRef.rnc = _.filter(individuRef.rnc, function(rnc) {
-                return rnc.montant === 0 || rnc.montant;
-            });
+                // Remove empty values from rnc
+                individuRef.rnc = _.filter(individuRef.rnc, function(rnc) {
+                    return rnc.montant === 0 || rnc.montant;
```
MattiSG: > A more readable version could be `rnc.hasOwnProperty('montant')`, or `'montant' in rnc`, or `typeof rnc.montant == 'number'`.

app/js/controllers/foyer/yearMoins2.js
```diff
@@ -36,27 +36,29 @@ angular.module('ddsApp').controller('FoyerRessourceYearMoins2Ctrl', function($sc
...
-                individuRef.individu.ressources.push({
-                    periode: $scope.yearMoins2,
-                    type: rnc.categorie.id,
-                    montant: rnc.montant
+                individuRef.rnc.forEach(function(rnc) {
+                    individuRef.individu.ressources.push({
+                        periode: $scope.yearMoins2,
+                        type: rnc.categorie.id,
+                        montant: rnc.montant
+                    });
```
MattiSG: > A better way to write this would be to `map` the `individuRef.rnc`, and then `concat` the result to `individuRef.individu.ressources`.

app/views/partials/foyer/ressources/year-moins-2.html
```diff
@@ -4,42 +4,57 @@
     Ces informations se trouvent sur votre déclaration de revenus d'impôts {{ yearMoins2 }}.
     <br>Vous pouvez la retrouver en ligne sur <a href="http://www.impots.gouv.fr/">impots.gouv.fr</a>.
   </p>
-  <form class="form-horizontal" novalidate ng-submit="submit()">
+  <form class="form-horizontal" novalidate name="ym2form" ng-submit="submit(ym2form)">
```
MattiSG: Why add a `name` to a `<form>`? This has been deprecated in HTML4, and is only used for navigating `document.forms`. Moreover, adding `form` in the name of a form seems redundant.

Considering how widely this is used throughout the app, I guess this is because Angular uses the name as a reference within the scope. Then, it would probably be wise to keep the same convention as most other forms and call it `form`, with a `ng-submit="submit(form)"` attribute 🙂

app/views/partials/foyer/ressources/year-moins-2.html
```diff
@@ -4,42 +4,57 @@
...
     <div class="row">
       <div class="col-md-12" ng-repeat="individuRef in individuRefsToDisplay">
         <h2>{{ individuRef.label }}</h2>
-        <div class="form-group" ng-repeat="rnc in individuRef.rnc">
-          <label class="control-label col-sm-3" for="{{ rnc.categorie.id }}-{{ individuRef.individu.role }}">
+        <div
+          class="form-group"
+          ng-repeat="rnc in individuRef.rnc"
+          ng-init="input_name = rnc.categorie.id +  '-' + individuRef.individu.role"
```
MattiSG: Why snake case?

app/views/partials/foyer/ressources/year-moins-2.html
```diff
@@ -4,42 +4,57 @@
...
                 class="form-control text-right"
                 ng-model="rnc.montant"
                 min="0"
                 select-on-focus
-                zero-to-empty>
+                zero-to-empty
+                aria-describedby="help-{{ input_name}}">
               <span class="input-group-addon hidden-xs">€</span>
             </span>
+            <ul class="help-block ym2-error">
```
MattiSG: Why introduce this `ym2-error` class? Didn't the global error handler work for displaying or hiding the `help-block`?

app/views/partials/foyer/ressources/year-moins-2.html
```diff
@@ -4,42 +4,57 @@
...
+            <p class="help-block" id="help-{{ input_name}}" ng-show ="{{ getDefaultValue(individuRef, rnc.categorie) }}">
               Ce montant vaut {{ getDefaultValue(individuRef, rnc.categorie) }}&nbsp;€ pour les 12 derniers mois.
             </p>
           </div>
         </div>
       </div>
       <div class="col-md-12" ng-repeat="individuRef in individuRefsToHide">
         <br/>
-        <button type="button"class="btn btn-sm btn-success col-md-6" ng-click="display(individuRef)">
+        <button type="button" class="btn btn-sm btn-success col-md-6" ng-click="display(individuRef)">
```
MattiSG: 👍

app/views/partials/foyer/ressources/year-moins-2.html
```diff
@@ -4,42 +4,57 @@
...
           Déclarer des ressources pour {{ individuRef.label }}
         </button>
       </div>
     </div>
     <div class="text-right">
+      <span class="text-danger" ng-if="ym2form.$submitted && ym2form.$invalid">
```
MattiSG: Why not use some `error` class here?

test/integration/components/YM2ResourcesComponent.js
```diff
@@ -1,6 +1,8 @@
 title: 'h1',
 demandeurRevenuActiviteInput: '#salaire_imposable_ym2-demandeur',
-suggestionText: '.help-block',
+suggestionText: 'p.help-block:nth-of-type(1)',
```
MattiSG: > This is equivalent to [`:first-of-type`](https://developer.mozilla.org/fr/docs/Web/CSS/:first-of-type) 🙂

app/views/partials/foyer/ressources/year-moins-2.html
```diff
@@ -4,42 +4,57 @@
...
-                id="{{ rnc.categorie.id }}-{{ individuRef.individu.role }}"
+                id="{{ input_name }}"
+                name="{{ input_name }}"
                 class="form-control text-right"
                 ng-model="rnc.montant"
                 min="0"
                 select-on-focus
-                zero-to-empty>
+                zero-to-empty
+                aria-describedby="help-{{ input_name}}">
```
MattiSG: Then you need to also add the `help-{{ input_name }}` ID to the matching help element :wink:

app/views/partials/foyer/ressources/year-moins-2.html
guillett: I wanted to use CSS to show and hide DOM elements but that was not a good idea (done differently everywhere else). I reverted that change in a `fixup` commit and used the _standard_ method (ie. `ng-if`/`ng-show`).

app/views/partials/foyer/ressources/year-moins-2.html
```diff
@@ -4,42 +4,57 @@
...
         <br/>
-        <button type="button"class="btn btn-sm btn-success col-md-6" ng-click="display(individuRef)">
+        <button type="button" class="btn btn-sm btn-success col-md-6" ng-click="display(individuRef)">
           Déclarer des ressources pour {{ individuRef.label }}
         </button>
       </div>
     </div>
     <div class="text-right">
+      <span class="text-danger" ng-if="form.$submitted && form.$invalid">
+        Un des montants indiqués est invalide
```
MattiSG: This is not very consistent with how we validate errors in other places of the app, where we only present the error where it happened. Plus, the rendering is not very nice. I'm not sure there's a lot of added value with this message.

capture d ecran 2017-06-27 a 18 07 49

app/views/partials/foyer/ressources/year-moins-2.html
```diff
@@ -4,42 +4,57 @@
...
                 select-on-focus
-                zero-to-empty>
+                zero-to-empty
+                aria-describedby="help-{{ inputName }}">
               <span class="input-group-addon hidden-xs">€</span>
             </span>
+            <ul class="text-danger" ng-if="form.$submitted && form[inputName].$invalid">
+              <li>
+                Montant invalide
+              </li>
```
MattiSG: This doesn't seem to use exactly the same structure as other validators, so I don't see the benefits of using the `<ul><li>` construct when there is only one element and the display is controlled only with Angular.

Removing the disc and aligning to the right would probably improve the aesthetics of the error. Also, I find it strange that the error is bigger than the label.

capture d ecran 2017-06-27 a 18 08 13

app/views/partials/foyer/ressources/year-moins-2.html
```diff
@@ -4,42 +4,57 @@
...
                 class="form-control text-right"
                 ng-model="rnc.montant"
                 min="0"
                 select-on-focus
-                zero-to-empty>
+                zero-to-empty
+                aria-describedby="help-{{ inputName }}">
               <span class="input-group-addon hidden-xs">€</span>
             </span>
+            <ul class="text-danger" ng-if="form.$submitted && form[inputName].$invalid">
```
MattiSG: Using `ng-show` rather than `ng-if` would prevent the whole layout bouncing around when errors appear  🙂

app/views/partials/foyer/ressources/year-moins-2.html
guillett: I agree the UI is not great but I thought [we agree that message is a good thing](https://github.com/sgmap/mes-aides-ui/pull/448#discussion_r103001295).

PR #610, opened 2017-07-07
Rely on id rather than _id for individuals

guillett:


guillett:
* UI tested locally
* OpenFisca request generation tested prior to migration
    * individu id are MongoDB.ObjectId (updated on refresh)
* OpenFisca request generation tested prior to migration
    * individu id are values previously set as _id (MongoDB.ObjetId or `demandeur`, `conjoint`, `enfant_X`)

PR #644, opened 2017-08-29
Make sure an annual 0 can be used as a YM2 proxy

guillett:


guillett:
Linked to https://github.com/openfisca/openfisca-core/issues/564

PR #707, opened 2017-11-20
Try to redirect users to partner websites

guillett:
* Create qualified flow
* Limit redirection work


guillett:
Contributes to #704.

PR #750, opened 2017-12-27
Remove deprecation warnings from Mongoose

guillett:


guillett:
`useMongoClient` option tested on production server with a local instance on port 9001. It is working as expected (`PORT=9001 node server.js`, `curl localhost:9001` and `curl localhost:9001/api/situations/53d24d39e40a7902009c3378`).

PR #827, opened 2018-05-04
A/B test underlined aid on result page

guillett:
People do not click on partner links.

We want to know if aid section headers should appear more click-able.


guillett:
Current view
capture d ecran de 2018-05-04 14-30-12


Alternative view
capture d ecran de 2018-05-04 14-29-16

PR #850, opened 2018-06-18
Affiche le montant de l'AAH dans le cadre de l'expérimentation

guillett:
Please @alexsegura have a look but do not merge.


app/styles/resultat.css
```diff
@@ -23,10 +23,6 @@
     font-size: 80%;
 }
 
-#simulationId {
```
alexsegura: It is also used in:
- `app/views/partials/foyer/recap-situation.html`
- `app/views/partials/trace.html`

app/js/services/resultatService.js
```diff
@@ -9,7 +9,8 @@ angular.module('ddsApp').service('ResultatService', function($http, droitsDescri
     *@return   {Object}  A new object containing the ressources of the family and of the individual. The family ressources will be overridden if conflicting.
     */
     function normalizeOpenfiscaRessources(testCase) {
-        return _.merge({}, testCase.menages._, testCase.familles._, testCase.individus.demandeur);
+        var individuId = testCase.menages._.personne_de_reference[0];
```
alexsegura: Is there always at least 1 element in the array?

guillett: Well, yes, in all valid situation. Worst case scenario, arror should be capture in the catch section.

alexsegura:
:shipit:

PR #857, opened 2018-06-26
Add trampoline for AAH test

guillett:


app/js/services/trampolineService.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+angular.module('ddsCommon').factory('TrampolineService', function($localStorage) {
```
alexsegura: I think we could avoid the ["smurf naming convention"](https://blog.codinghorror.com/new-programming-jargon/)  😀

A cleaner interface would be: 
- `set(name, value = null)` → accepts a hash or a name + value
- `get(name)` → returns the whole hash, or just the key if provided. When a key is retrieved, it is removed 
- `clear()` → clears everything
- `has(name)` → allows to check if a key exists, without deleting it 
- `keys()` → returns the keys, without deleting them

app/js/directives/droitsEligiblesList.js
```diff
@@ -12,13 +12,15 @@ angular.module('ddsApp').directive('droitEligiblesList', function() {
...
 
-angular.module('ddsApp').controller('droitsEligiblesListCtrl', function($scope) {
+angular.module('ddsApp').controller('droitsEligiblesListCtrl', function($scope, TrampolineService) {
     $scope.isNumber = _.isNumber;
     $scope.isString = _.isString;
     $scope.shouldDisplayYM2Warning = function(droit) {
         return droit.isBaseRessourcesYearMoins2 && ! $scope.ressourcesYearMoins2Captured && ! _.isString(droit.montant);
     };
 
+    $scope.setTrampoline = TrampolineService.setTrampoline;
```
alexsegura: Why not: 

```
$scope.trampoline = TrampolineService;
```

app/js/services/trampolineService.js
alexsegura: Inspired by [Symfony Flash Messages](https://symfony.com/doc/current/components/http_foundation/sessions.html#flash-messages)  😎

guillett:
Many thanks @alexsegura. A second look ? :)

PR #858, opened 2018-06-26
Fix undefined error on result page

guillett:
* AB testing environment may not be set


app/js/controllers/resultat.js
```diff
@@ -7,7 +7,7 @@ angular.module('ddsApp').controller('ResultatCtrl', function($scope, $rootScope,
     $scope.warningMessage = false;
 
     var env = ABTestingService.getABTestingEnvironment();
-    $scope.linkAlternative = (env.link && env.link.value) || 'B';
+    $scope.linkAlternative = (env && env.link && env.link.value) || 'B';
```
guillett: `env` may be null, this implementation avoids an error.

PR #860, opened 2018-06-26
Actually make watch for Express files work.

alexsegura:
Since e554c6fc3afcb9c275c0b9da2b6063bdc689b378, it seems that JSHint is not managed via Grunt anymore. 

However, `newer:jshint:node` is still called when a file changes for Express stuff, causing an error: 

```
Running "newer:jshint:node" (newer) task
Warning: Cannot read property 'files' of undefined
```

I removed the task. I think we can get rid of `grunt-newer` as well. 

Also, I added the files in `backend/**/*.js` to the watch list, so that the server is restarted when there  are changes in the controllers, for example.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -78,9 +78,10 @@ module.exports = function (grunt) {
       express: {
         files: [
           'server.js',
-          'index.js'
+          'index.js',
+          'backend/**/*.js'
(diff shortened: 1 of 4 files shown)


alexsegura:
@guillett :shipit:

PR #940, opened 2018-10-17
Remove prestation data duplication in ludwig configuration

guillett:
* Rely on droits.js

### DIFF
--- a/backend/lib/ludwig/api-possible-values.js
+++ b/backend/lib/ludwig/api-possible-values.js
@@ -1,122 +1,38 @@
-module.exports = [
-    {
-        'id':'cmu_c',
-        'hasMontant':false,
-        'shortLabel':'CMU-C',
-    },
-    {
-        'id':'acs',
-        'shortLabel':'ACS',
-    },
-    {
-        'hasMontant':false,
-        'id': 'apa_eligibilite',
-        'shortLabel': 'APA - Éligibilité',
-    },
-    {
-        'id':'aspa',
-        'shortLabel':'ASPA',
-    },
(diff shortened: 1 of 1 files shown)


backend/lib/ludwig/api-possible-values.js
```diff
@@ -1,122 +1,38 @@
...
+var possibleValues = [];
+Object.keys(droits).forEach(function(prestationLevel) {
+    var providers = droits[prestationLevel];
+
+    Object.keys(providers).forEach(function(providerName) {
+        var provider = providers[providerName];
+
+        var options = Object.keys(provider.prestations).map(function(prestationName) {
+            var prestation = provider.prestations[prestationName];
+            var label = prestation.label == 'Tarification solidaire transports' ? (provider.label + ' - ' + prestation.label) : prestation.label;
```
alexsegura: Can we avoid hardcoding this? I think it is ok to append the city name to all local benefits.

guillett: Thanks, I refactored that section.

backend/lib/ludwig/api-possible-values.js
```diff
@@ -1,122 +1,50 @@
...
-    },
-    {
-        'id': 'brest_metropole_transport',
-        'shortLabel': 'Brest - Transport',
-        'unit': '%',
-    },
-];
+var droits = require('../../../app/js/constants/droits');
+
+// Mitigate label duplicates by counting labels and appending provider label with necessary
```
alexsegura: 👌

PR #941, opened 2018-10-17
Use custom Docker image with Node & Python.

alexsegura:
This is a first version. To save a few seconds, I still need to install pip directly on the Docker image (pip is not installed by default on Python < `2.7.9`)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,15 +6,28 @@ defaults: &defaults
   working_directory: ~/mes-aides-ui
   docker:
     # https://circleci.com/docs/2.0/circleci-images/#language-image-variants
-    - image: circleci/node:6.14.1-browsers
+    - image: betagouv/mes-aides-docker:latest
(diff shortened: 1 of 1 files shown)


guillett:
Cool, can you drop the submodule in that PR to limit the complexity added to the CI configuration ?

alexsegura:
@guillett I would prefer to create a new PR for this, I need to experiment with Git commands to merge the submodule back in the main repo while preserving history.

guillett:
Can you confirm that the docker image `betagouv/mes-aides-docker:latest` won't be overriden as commits are pushed on `https://github.com/betagouv/mes-aides-openfisca` ?

alexsegura:
No, it is not the same image. 

- https://hub.docker.com/r/betagouv/mes-aides-openfisca/builds/
- https://hub.docker.com/r/betagouv/mes-aides-docker/builds/

PR #945, opened 2018-10-29
Prevent duplicates in injected aids

guillett:
The current OpenFisca payload processing leads to duplicates in injected aids.

With this PR I push forward the previous implementation of ResultatService and add test to ensure, that local aids and visible and that no duplicates are generated.

This is a naive implementation but the result page is currently unusable when injected aids are shown.


guillett:
Testé en local, build de dev et de production.

PR #950, opened 2018-11-05
Add tests for YAML test generation

guillett:
We want to move away from [Ludwig v1](https://github.com/betagouv/ludwig-ui) and use [Ludwig v2](https://github.com/betagouv/ludwig).

Ludwig 1 was two things at the same time:
- an expectation editor,
- a test directory and
- a test runner

With Ludwig 1 it was easy to generate tests but they were hard to use for developers (tests are in a database). 

Ludwig 2 is a way to link applications to their repository and their test base.
The Ludwig 2 v0 (🙃) is a single endpoint to POST file test content to be added to the underlying repository.

This PR ensure that generated YAML are valid OpenFisca tests.

### DIFF
--- a/test/backend/openfisca/test.js
+++ b/test/backend/openfisca/test.js
@@ -1,4 +1,5 @@
 var expect = require('expect');
+var _ = require('lodash');
 var fs = require('fs');
 var subject = require('../../../backend/lib/openfisca/test');
 var tmp = require('tmp');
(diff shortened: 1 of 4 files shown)


app/js/services/situationService.js
```diff
@@ -115,8 +115,12 @@ angular.module('ddsCommon').factory('SituationService', function($http, $session
             situation.individus.forEach(function(individu) {
                 individu.date_naissance = individu.date_naissance.format('YYYY-MM-DD');
                 delete individu.hasRessources;
+
+                if (individu.dateDernierContratTravail) {
```
alexsegura: Is this related to this Pull Request?

test/backend/openfisca/test.js
```diff
@@ -40,4 +80,24 @@ describe('openfisca generateYAMLTest', function() {
     it('contains provided output_variables', function() {
         expect(result).toInclude('valueOne: 1');
     });
+
+    if (process.env.INTEGRATION_TEST) {
```
alexsegura: Can't we use something else to make sure we are in a virtual env? 
Maybe `VIRTUAL_ENV` variable?

test/backend/openfisca/test.js
```diff
@@ -30,6 +39,38 @@ describe('openfisca generateTest', function() {
...
+    var respErr = "";
+    var respOut = "";
+
+    child.stdout.on('data', function (buffer) { respOut += buffer.toString(); });
+    child.stderr.on('data', function (buffer) { respErr += buffer.toString(); });
+    child.on('exit', function(code) { callback(code ? 'Exit code was ' + code : null, respOut, respErr); });
+    child.on('error', function(err) { callback(err); });
+}
+
+function runOpenFiscaTest(yaml, extension, done) {
```
alexsegura: IMHO this should be in `backend/lib/openfisca/test.js`, and this shouldn't make any assertion. 
Assertions should be move inside the test (i.e, assert this command has returned zero exit code).

alexsegura: Maybe we could return a Promise instead of using callback?

```
runOpenFiscaTest(yaml, extension)
  .then(stdout => console.log(stdout))
  .fail(e => console.log(e.code))
```

PR #951, opened 2018-11-05
Improve YAML representation

guillett:
Previously, too many attributes were displayed and `dateDernierContratTravail` is ugly (a moment instead of a YYY-MM-DDD date).

capture d ecran de 2018-11-05 13-35-00


app/js/services/situationService.js
```diff
@@ -115,8 +115,12 @@ angular.module('ddsCommon').factory('SituationService', function($http, $session
             situation.individus.forEach(function(individu) {
                 individu.date_naissance = individu.date_naissance.format('YYYY-MM-DD');
                 delete individu.hasRessources;
+
+                if (individu.dateDernierContratTravail) {
```
alexsegura: Maybe `jsyaml` has a way to always serialize instances of `Moment.js` into a certain format?

PR #953, opened 2018-11-05
Export tests to appropriate repositories

guillett:


app/js/controllers/foyer/suggestion.js
```diff
@@ -79,31 +67,71 @@ angular.module('ddsApp').controller('SuggestionCtrl', function($scope, $http,dro
         delete expectedResult.shouldCompute;
     };
 
+    $scope.determineExtensionAndRepository = function(aids) {
```
alexsegura: I don't see no particular reason for binding this to `$scope`, it looks like it is like this just to make it exposed to tests. If this is the reason, then I think this should be extracted to a `SuggestionService`

app/js/controllers/foyer/suggestion.js
```diff
@@ -79,31 +67,71 @@ angular.module('ddsApp').controller('SuggestionCtrl', function($scope, $http,dro
...
+        }
+
+        var dest = 'openfisca-' + extensions[0];
+        return {
+            extension: dest,
+            repository: dest,
+        };
+    };
+
+    $scope.generateTestMetadata = function(test, extension) {
```
alexsegura: Same comment about binding to `$scope`. This should be a simple function.

Issue #888, opened 2018-08-14
Remove bluebird ?


alexsegura:
There is a problem when running a fresh install of Mes Aides. 

```
TypeError: model.create(...).then(...).end is not a function
    at /srv/app/node_modules/ludwig-api/fixtures/index.js:31:16
    at Query.<anonymous> (/srv/app/node_modules/mongoose/lib/model.js:4081:16)
    at /srv/app/node_modules/kareem/index.js:273:21
    at /srv/app/node_modules/kareem/index.js:131:16
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
    at process._tickCallback (internal/process/next_tick.js:104:9)
```

It can be fixed by commenting the [following line](https://github.com/betagouv/mes-aides-ui/blob/dc4aea26c6875ade9b34287ff8bbf9ecb2dfa42b/backend/config/mongoose.js#L6) in `backend/config/mongoose.js`

```diff
var path = require('path');
var fs = require('fs');

module.exports = function(mongoose, config) {

-    mongoose.Promise = require('bluebird');

    mongoose.connect(config.mongo.uri, config.mongo.options);

    // Bootstrap models
    var modelsPath = path.join(__dirname, '../models');
    fs.readdirSync(modelsPath).forEach(function (file) {
        if (/(.*)\.(js$|coffee$)/.test(file)) {
            require(modelsPath + '/' + file);
        }
    });

};
```

- Is there any significant benefit in using bluebird? 
- Is this mandatory with Mongoose?

From what I see, we are just using the `promisifyAll` utility.

guillett:
Mongoose require a promise library. Maybe It was possible to do `mongoose.Promise = Promise`.

Plus `promisifyAll` is a good utility function.

The issue with `ludwig` was fixed. I think Bluebird can be kept. I close this issue.

PR #962, opened 2018-11-22
Rely on fixed ludwig-api

guillett:
I don't think removing bluebird is the way to go. I have done a hotfix in Ludwig API.

closes #927  and #888

`end()` is not mandatory cf.
- https://mongoosejs.com/docs/4.x/docs/api.html#promise_Promise-end
- https://github.com/aheckmann/mpromise#end

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "jsonwebtoken": "^8.3.0",
     "kerberos": "0.0.23",
     "lodash": "^4.17.4",
-    "ludwig-api": "^1.2.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "jsonwebtoken": "^8.3.0",
     "kerberos": "0.0.23",
     "lodash": "^4.17.4",
-    "ludwig-api": "^1.2.0",
+    "ludwig-api": "betagouv/ludwig-api#rm-invalid-end",
```
alexsegura: Would be cleaner to use a tag. `1.3.0`?

PR #1005, opened 2018-12-27
Add Sentry in backend

guillett:


backend/config/config.js
```diff
@@ -10,14 +10,20 @@ var all = {
             useMongoClient: true,
         },
     },
-    sessionSecret: process.env.SESSION_SECRET || 'fghjdfjkdf785a-jreu'
+    sessionSecret: process.env.SESSION_SECRET || 'fghjdfjkdf785a-jreu',
+    sentry: require('./sentry'),
```
alexsegura: I think this does not make sense to store objects in the config, it should stay a pure JavaScript object (i.e, serializable with `JSON.stringify`)

backend/config/sentry.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
```
alexsegura: I think it would be simpler to have everything in only one file. Just store this under a `sentry` key in the main config?

backend/config/config.js
```diff
@@ -10,14 +10,20 @@ var all = {
...
 };
 
 var override = {};
 try
 {
     override = require('./' + env);
     console.info('Using specific configuration for ' + env + '.');
+
+    if (override && override.sentry && override.sentry.privateDsn) {
+        override.sentry.instance = require('raven');
```
alexsegura: Why do we need this?

alexsegura:
@guillett Niiiiiice 🐛

backend/config/sentry.js
guillett: The main configuration file should not be imported in a public file otherwise all our secrets will be made public.

backend/config/config.js
guillett: That's the case.

alexsegura: Oh, my bad, you are right 🙂

backend/config/sentry.js
alexsegura: They will be made public if someone does `window.config = ...` but otherwise it won't, no? 
Everything is wrapped in anonymous functions with Webpack

guillett: Everything in `backend/config/sentry.js` is made available from the `dist` folder. If I import the main configuration file (which contains a Sentry DSN with a secret) that secret will be made public.

I may not understand your proposal correctly. Let's chat away from keyboard if that's the case. :smiley:

PR #1502, opened 2021-03-15
Update install instructions and metadata

MattiSG:


guillett (approved):
LGTM. Merci @MattiSG !

PR #1743, opened 2021-08-04
fix les packages

Kout95:


Kout95:
Package requis : `sudo apt-get install libkrb5-dev`

QuentinMadura:
Close PR with #1789

PR #3138, opened 2022-08-16
Permet le lancement manuel de la vérification des liens

guillett:
https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-15


Kout95:
Très bonne idée !

